PR #209, opened 2021-03-05
Check that GUIDs are unique

iwsmith:
Check configs to make sure GUIDs used by Slates and SlateLineups are unique.

## Update
Based on PR feedback I did the following:
* Moved slate and lineup validation code into their respective models. This means we do the validations on model load, which is probably right so we don't end up with invalid models being loaded
* Cross GUID checks (between slate and lineup) done at app start.
* Tests

### DIFF
--- a/tests/functional/json/slate_lineup_config.py
+++ b/tests/functional/json/slate_lineup_config.py
@@ -1,15 +1,27 @@
 from app.models.slate_lineup_experiment import SlateLineupExperimentModel
 from app.models.slate_config import SlateConfigModel
-from app.models.slate_lineup_config import SlateLineupConfigModel
+from app.models.slate_lineup_config import SlateLineupConfigModel, validate_unique_guids
+import pytest
+
+
+def test_validate_config():
+    # Overlapping GUIDs between configs
+    slate_config = [SlateConfigModel("617d9613-d338-4cc0-92ba-9a90831f6a46", "dupe slate", "dupe slate")]
+    lineup_config = [SlateLineupConfigModel("617d9613-d338-4cc0-92ba-9a90831f6a46", "Duplicated GUID")]
+    with pytest.raises(ValueError):
+        validate_unique_guids(lineup_config, slate_config)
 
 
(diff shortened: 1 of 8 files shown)


tests/functional/json/slate_lineup_config.py
```diff
@@ -1,15 +1,32 @@
...
+            raise ValueError(f"Slate GUID {sid} appears more than once in slate config")
+
+    lineup_ids = Counter([r.id for r in SlateLineupConfigModel.SLATE_LINEUP_CONFIGS])
+    for lid, count in lineup_ids.items():
+        if count > 1:
+            raise ValueError(f"SlateLineup GUID {lid} appears more than once in slate lineup config")
+
+    guid_overlap = set(lineup_ids.keys()).intersection(slate_ids.keys())
+    if guid_overlap:
+        raise ValueError(f"GUID is reused for slate and lineup: {guid_overlap}")
```
chelseatroy, line 27: There are faster ways to do this, but they either require an extra step to get the actual duplicated ID or aren't as clear to follow as this, so I'm happy with this solution.

chelseatroy (changes requested):
The solution I'm happy with. I think it needs three automated tests:

- one for duped slate ids
- one for duped layout ids
- one for reused slate ids among the layout ids

jpetto (approved):
looks solid to me 🙌

kkyeboah (commented):
This works for me but deferring to @chelseatroy for the requested changes.

iwsmith:
> The solution I'm happy with. I think it needs three automated tests:
> 
>     * one for duped slate ids
> 
>     * one for duped layout ids
> 
>     * one for reused slate ids among the layout ids
> 
> 
> Happy to pair on that if that's helpful!

Hmm, I've never written a test for a test before. I think this really indicates this code shouldn't be in a unit test, but instead somewhere else. I will move stuff around and resubmit.

chelseatroy (approved):
That's the stuff!

PR #214, opened 2021-03-12
[SHIPIT-123] Fix topic slate order

mmiermans:
# Goal
The Web repo assumes that topic linups have a curated and algorithmic slate, in this order. Whether clients can depend on slates coming in a particular order is a great question, that we should talk about.

### DIFF
--- a/app/json/slate_lineup_configs.json
+++ b/app/json/slate_lineup_configs.json
@@ -22,7 +22,7 @@
     "experiments": [
       {
         "description": "SlateLineup test 1",
-        "rankers": [ ],
+        "rankers": [],
         "slates": [
           "de254c5d-57a7-4553-850f-153ee385014d",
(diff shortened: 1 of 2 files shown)


app/json/slate_lineup_configs.json
```diff
@@ -22,7 +22,7 @@
     "experiments": [
       {
         "description": "SlateLineup test 1",
-        "rankers": [ ],
```
mmiermans, line 25: I regenerated this file to correct the slate order, so select 'Hide whitespace changes' to see only the relevant changes.

chelseatroy (approved):
Yep—this is really brittle right now because in the Web repo we just have to hope the topic pages come back with two slates, curated first, followed by algorithmic. Ideally there'd be some kind of labels I think on the slates, so that clients don't just have to trust that they're in the expected order.

PR #218, opened 2021-03-17
Add RecIt CandidateSet Support

iwsmith:
# Goal
This adds a new type of CandidateSet, a `RecItCandidateSet`, which allows us to make slates that use RecIt as the source of items. This specifically allows for personalized slates using the bestof, curated, or syndicated corpora. Once the vector store exists we would move this functionality there, but there isn't any timeline on that work currently. This allows us to test out the performance of personalized slates sooner.

See the [design doc](https://docs.google.com/document/d/14PaPC3xtG7ajdH7U5VguO3J-ePoUIp5Egm2EBkWYd-M/edit) for more details.

# Notes
* Created a `DynamoDBCandidateSet` for the existing `CandidateSetModel` functionality
* You can now specify a `RecItCandidateSet` as `recit-personalized/bestof` (or `curated` or `syndicated`) in a slate.
* If a user doesn't have a profile we return an empty CandidateSet object. We either need to filter this out of the response, or clients will need to. We haven't really made a decision about that yet.
* Added a `candidate_set_factory` that will use the appropriate candidate set type based on the candidate set id

### DIFF
--- a/.docker/localstack/dynamodb/recommendation_api_candidate_sets_data.json
+++ b/.docker/localstack/dynamodb/recommendation_api_candidate_sets_data.json
@@ -3,44 +3,137 @@
     {
       "PutRequest": {
         "Item": {
+          "id": {
+            "S": "f340f4ad-ce0a-543a-a87f-ca8321a3c978"
+          },
+          "version": {
+            "N": "1"
+          },
+          "flow": {
+            "S": "CollectionsFlow"
+          },
+          "created_at": {
+            "N": "1615335211"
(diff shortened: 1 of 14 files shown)


mmiermans (commented):
We paired on this.

I believe it needs a load test, but that we can only do that in production due to missing data in Pocket-Dev. (Correct me if I'm wrong, Ian.)

kkyeboah (approved):
lgtm 👍 Agree on the load test.

PR #225, opened 2021-03-22
[BACK-680] Switch the format of the click data input to the new source

mmiermans:
# Goal
Get clickdata by slate id, instead of using topic-based clickdata by default.

Tickets:
* https://getpocket.atlassian.net/browse/BACK-680

## Implementation Decisions
- This is based on https://github.com/Pocket/recommendation-api/pull/221, but we wanted to split in two:
    1. Get clickdata by slate_id (this PR)
    2. Remove the old topics graphQL endpoint (blocked by the Web repo removing references)

### DIFF
--- a/app/models/clickdata.py
+++ b/app/models/clickdata.py
@@ -11,19 +11,14 @@ import app.cache
 
 # batch get has a 100 item limit
 # https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html
-def chunks(index, n=100):
+def _chunks(index, n=100):
     """Yield successive n-sized chunks from index."""
     for i in range(0, len(index), n):
         yield index[i: i + n]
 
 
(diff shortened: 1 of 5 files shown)


app/models/clickdata.py
```diff
@@ -11,19 +11,14 @@
 
 # batch get has a 100 item limit
 # https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html
-def chunks(index, n=100):
+def _chunks(index, n=100):
```
mmiermans, line 14: Renamed according to Python convention because these functions are intended to be kept private.

mmiermans:
> Makes sense to me. I assume the clickdata has been updated accordingly?

No, but Matt said it was fine if we temporarily don't do Thompson sampling on the explore pages.

PR #226, opened 2021-03-23
[BACK-683] Dedupe recs

mmiermans:
# Goal
Deduplicate recommendations across slates in a lineup. If a recommendation occurs more than once, it's removed from all but the first slate.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/BACK-683

## Implementation Decisions
- test_slate_lineup_model is almost at a point where a fixture library would be convenient, but I think it's still manageable for now.

### DIFF
--- a/app/models/slate.py
+++ b/app/models/slate.py
@@ -104,3 +104,15 @@ class SlateModel(BaseModel):
             recommendations=recommendations,
             requestId=str(uuid.uuid4()),
         )
+
+
+async def deduplicate_recommendations_across_slates(slates: List[SlateModel]) -> List[SlateModel]:
+    seen_item_ids = set()
+
+    for slate in slates:
+        # Remove recommendations that exist in previous slates
+        slate.recommendations = [r for r in slate.recommendations if r.item_id not in seen_item_ids]
+        # Add all item ids from slate to seen_item_ids
+        seen_item_ids |= {r.item_id for r in slate.recommendations}
+
+    return slates
--- a/tests/functional/models/test_slate_lineup_model.py
+++ b/tests/functional/models/test_slate_lineup_model.py
@@ -8,14 +8,34 @@ from app.models.slate_experiment import SlateExperimentModel
 from app.models.slate_lineup_experiment import SlateLineupExperimentModel
 from app.models.slate_lineup_config import SlateLineupConfigModel
 
-slate_lineup_config_id = 'test-slate_lineup-config-id'
-slate_lineup_experiment = SlateLineupExperimentModel('test-ex', 'test-ex-desc', ['top15'], ['test-slate-id'])
-slate_lineup_config_model = SlateLineupConfigModel(slate_lineup_config_id, 'test-desc', [slate_lineup_experiment])
-
-slate_config_id = 'test-slate_lineup-config-id'
+# First slate
+slate_config_id = 'test-slate-config-id'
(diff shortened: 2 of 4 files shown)


app/models/slate.py
```diff
@@ -104,3 +104,15 @@ async def __get_slate_from_slate_config(slate_config: 'SlateConfigModel',
...
+async def deduplicate_recommendations_across_slates(slates: List[SlateModel]) -> List[SlateModel]:
+    seen_item_ids = set()
+
+    for slate in slates:
+        # Remove recommendations that exist in previous slates
+        slate.recommendations = [r for r in slate.recommendations if r.item_id not in seen_item_ids]
+        # Add all item ids from slate to seen_item_ids
+        seen_item_ids |= {r.item_id for r in slate.recommendations}
+
+    return slates
```
chelseatroy, line 118: 🦢

tests/functional/models/test_slate_lineup_model.py
```diff
@@ -90,3 +110,50 @@ async def test_get_slate_lineup_no_slates(self, slate_config, slate_lineup_confi
 
         slate_lineup = await SlateLineupModel.get_slate_lineup(slate_lineup_config_id, slate_count=0)
         assert len(slate_lineup.slates) == 0
+
+    @patch.object(SlateConfigModel, 'SLATE_CONFIGS_BY_ID', slate_configs_by_id)
+    @patch('app.models.slate_lineup_config.SlateLineupConfigModel.find_by_id', return_value=slate_lineup_config_model_2)
+    async def test_get_slate_lineup_dedupe_recs(self, slate_lineup_config):
```
chelseatroy, line 116: We're doing a weird combination of patching and seeding in these tests. We've been doing it since before this PR so I don't think we should hold this PR on it, but might be worth a team chat about why we've alighted on this pattern and whether we're doing it because we _want_ to or we're doing it because something is _forcing_ us to that we might be able to fix

tests/unit/models/test_slate_model.py
```diff
@@ -0,0 +1,53 @@
...
+        assert deduped_slate_0.recommendations[0].item_id == '1'
+        assert deduped_slate_0.recommendations[1].item_id == '2'
+        assert deduped_slate_1.recommendations[0].item_id == '3'
+        assert deduped_slate_2.recommendations[0].item_id == '4'
+
+    def __get_slate_model(self, item_ids: List[str]) -> SlateModel:
+        recommendations = [
+            RecommendationModel.candidate_dict_to_recommendation({'item_id': item_id}) for item_id in item_ids
+        ]
+        return SlateModel(id=1, recommendations=recommendations)
```
chelseatroy, line 53: nice work on these unit tests 😎

chelseatroy (approved):
Ope, failed CI. I'll pull it down and fix the test!

chelseatroy (dismissed):
🍐

tests/functional/models/test_slate_lineup_model.py
mmiermans: Agreed. I think it would be nicer if we called `SlateConfigModel.load_slate_configs` before the test, and load it with a json fixture from tests/assets/json.

bassrock: Retro topic? I know we have a set of standards in TS but we have to talk about python or even set up a python linter.

chelseatroy (approved):
Re-signed and force pushed

bassrock (approved):

chelseatroy:
Oh dingdangit, hold on

PR #236, opened 2021-04-08
Don't use invalid MLE priors for modules

iwsmith:
For some reason we are generating negative parameters for our beta distribution for module priors. This is not valid. If we find invalid module level priors just use the naive default.


chelseatroy (approved):
Paired on this change!

PR #237, opened 2021-04-09
[SHIPIT-123] Fix explore homepage lineup

mmiermans:
# Goal
The Web client isn't getting enough recommendations. It needs at least 20. We accomplish this in two ways:
1. Limit curated nonsyndicated to 5, to prevent too many duplicates from being removed from the second slate.
2. Remove the top15 experiment from 'All Curated New Tab', so it can reliably return at least 15 recommendations.

## Reference
Slack thread:
- https://pocket.slack.com/archives/CR6T8BNCT/p1617924164014300

## Implementation Decisions
We created a slate 48e766be-5e96-46fb-acbf-55fee3ae8a28 based on 2e3ddc90-8def-46d7-b85f-da7525c66fb1 because Ian said the latter is used elsewhere, where we don't want to limit to 5 recommendations.

### DIFF
--- a/app/json/slate_configs.json
+++ b/app/json/slate_configs.json
@@ -58,6 +58,37 @@
     ]
   },
   {
+    "id": "48e766be-5e96-46fb-acbf-55fee3ae8a28",
+    "displayName": "Editors' Picks",
+    "description": "Top 5 curated items without syndicated that are at most a week old",
+    "experiments": [
+      {
+        "description": "default",
+        "candidateSets": [
+          "493a5556-9800-449f-8f8c-c27bb6c8c810"
+        ],
+        "rankers": [
+          "top30",
+          "thompson-sampling",
+          "pubspread",
+          "top5"
+        ]
+      },
+      {
+        "description": "ts15",
+        "candidateSets": [
+          "493a5556-9800-449f-8f8c-c27bb6c8c810"
+        ],
+        "rankers": [
+          "top15",
+          "thompson-sampling",
+          "pubspread",
+          "top5"
+        ]
+      }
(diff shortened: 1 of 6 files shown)


app/json/slate_configs.json
```diff
@@ -58,29 +58,49 @@
...
-    "description": "Curated items with syndicated that are at most a week old",
+    "id": "48e766be-5e96-46fb-acbf-55fee3ae8a28",
+    "displayName": "Editors' Picks",
+    "description": "Top 5 curated items without syndicated that are at most a week old",
     "experiments": [
       {
         "description": "default",
         "candidateSets": [
-          "35018233-48cd-4ec4-bcfd-7b1b1ccf30de"
+          "493a5556-9800-449f-8f8c-c27bb6c8c810"
```
bassrock, line 68: Do we need 2 experiments now?

bassrock (approved):
This looks right to me. That's 1 confusing diff though.

iwsmith (approved):

app/json/slate_configs.json
mmiermans: We didn't talk about changing this, and they're subtlety different when combined with Thompson sampling. I don't think it matters for our purposes, because we limit it to 5 in the end.

jeshuaborges (commented):
Nice work structuring the application. Making this update is easy for me to follow even though I have almost no exposure to this project. I think that is mostly a testament to the architecture of the application.

mmiermans:
Successfully QA'ed in Pocket-Dev:
Screenshot from 2021-04-09 10-40-35

PR #238, opened 2021-04-19
removing thompon sampling from curated topic slates

lfishhead:
# Goal

Editorial team would like more control over content of curated section of topic pages during tests with new tab in Q2.

Removing Thompson sampling from the curated sections of the explore topic pages will revert to approval based ordering and allow them to provide excerpts in the tiles on new tab that correspond predictably to the items appearing in the curated section of those pages.

No changes to algorithmic slates.

## Reference

Slack thread: https://pocket.slack.com/archives/CN0L1NDD3/p1618511471008400

## Implementation Decisions

Removed Thompson sampling ranker from curated topic slates.  
Alternative would be to set up separate curated topic slates with TS enabled.

### DIFF
--- a/app/json/slate_configs.json
+++ b/app/json/slate_configs.json
@@ -170,7 +170,6 @@
         ],
         "rankers": [
           "top30",
-          "thompson-sampling",
           "pubspread"
         ]
       }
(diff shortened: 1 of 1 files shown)


kkyeboah (approved):
👍

chelseatroy (approved):
LGTM!

PR #245, opened 2021-05-04
Fix alarm action for http request count

mmiermans:
# Goal
Update pocket/terraform-modules to 1.19.1 to pull in [a fix for the request count alarm action](https://github.com/Pocket/terraform-modules/commit/27c33b53f3b9f624bf41013176372290ef58de6d#diff-0731374296d9e4f7cac42a725e1b58a908e207fd57df2893966c30f7379b0592R849-R850), to make the HttpRequestCount alarm non-critical.

Additionally these changes were made:
- Update portMappings because of a breaking interface change in https://github.com/Pocket/terraform-modules/pull/132.
- Update Elasticache because of a breaking interface change in https://github.com/Pocket/terraform-modules/pull/122.
- Update cdktf and constructs to [match the version used in @pocket/terraform-modules](https://github.com/Pocket/terraform-modules/blob/main/package.json#L41-L42). No impact expected in production, but installing multiple versions of cdktf caused problems locally in the past.

## Reference

Slack thread:
* https://pocket.slack.com/archives/C016NAMSW2J/p1620122866003600

### DIFF
--- a/.aws/package-lock.json
+++ b/.aws/package-lock.json
@@ -1,3224 +1,74 @@
 {
   "name": "recommendation-api",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "recommendation-api",
-      "version": "1.0.0",
-      "license": "MPL-2.0",
-      "dependencies": {
-        "@pocket/terraform-modules": "1.9.0",
-        "cdktf": "0.1.0",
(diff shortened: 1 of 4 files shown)


kkyeboah (commented):
Codebuild error suggests we need updates to the elasticache service constructs as well. https://github.com/Pocket/recommendation-api/blob/main/.aws/src/elasticache.ts#L25

kkyeboah (approved):
👍

PR #255, opened 2021-05-12
switching display_name and description

lfishhead:
# Goal

Mathijs noticed  the `display_name` field in the slate config json file was being largely switched with the `description` field.  `display_name` should have text that can be displayed by clients to describe the slate, while the `description` is generally for internal consumption.  I tried to get these sorted out.

## Implementation Decisions

reading a long JSON file.


app/json/slate_configs.json
```diff
@@ -937,8 +937,8 @@
   },
   {
     "id": "0f322865-64e6-472d-8147-b3d6637a7d67",
-    "displayName": "Pocket Collections",
-    "description": "Curated by our editors Stories to fuel your mind",
+    "displayName": "Curated collections by our editors Stories to fuel your mind",
```
mmiermans, line 940: Grammar isn't right here.

app/json/slate_configs.json
```diff
@@ -778,8 +778,8 @@
   },
   {
     "id": "8a495385-9972-4d0e-b93e-4234f4897939",
-    "displayName": "Algorithmic Sports Slate",
-    "description": "Popular with Pocket readers  Stories from across the web",
+    "displayName": "Popular with Pocket readers  Stories from across the web",
```
mmiermans, line 781: This seems like two sentences are joined together

mmiermans (commented):
A couple display names don't sound right:
- Popular with Pocket readers  Stories from across the web
- Curated by our editors Stories to fuel your mind
- Curated collections by our editors Stories to fuel your mind

Thinking long-term, it would probably be nice to have this copy be managed by product, instead of by D&L.

app/json/slate_configs.json
lfishhead: it's literally what is on the topic pages now.  In other words, we are managing it, but someone else created it.

collectedmind:
```
      <SectionHeader
        sectionTitle={t('discover:curated-by-our-editors', 'Curated by our editors')}
        sectionDescription={t('discover:stories-to-fuel-your-mind', 'Stories to fuel your mind')}
      />
```

It is hardcoded at the moment, but would be awesome if it came down so curation could tweak it as will.

mmiermans: It sounds like we should to add a title property to slates.

We also didn't get around to localizing these strings. Maybe we can tackle both of those at once when we're doing another recommendation api project.

lfishhead: Am I wrong it would still be worth making this switch in the short term @collectedmind ?  If so, I will update this PR with an additional small lineup change we discussed this morning @mmiermans and re-request review.

app/json/slate_configs.json
```diff
@@ -289,9 +289,9 @@
   },
   {
     "id": "b0de37c0-81ef-4063-a432-1bb64270b039",
-    "displayName": "Curated Education Slate",
+    "displayName": "Curated by our editors Stories to fuel your mind",
```
iwsmith, line 292: Should these topic slates include the name of the slate somewhere?

app/json/slate_configs.json
```diff
@@ -1003,8 +1003,8 @@
   },
   {
     "id": "4d95bade-8a43-4700-9879-3a73ba1da640",
-    "displayName": "10 Incredible Long-Reads That’ll Transport You to Another Time and Place",
-    "description": "There’s nothing like getting lost in a fascinating story — and now is a better time than any to dive into another world.",
+    "displayName": "There’s nothing like getting lost in a fascinating story — and now is a better time than any to dive into another world.",
```
iwsmith, line 1006: Which client is using this currently?

bassrock (approved):

bassrock:
Per slack, approved and @lfishhead will follow up with names from carolyn when she is back from PTO. https://pocket.slack.com/archives/C01CKCB9746/p1624486510046100?thread_ts=1624485875.045300&cid=C01CKCB9746

PR #264, opened 2021-05-25
Adding a personalized topic slate lineup

lfishhead:
# Goal

The goal is to add a slate lineup that includes all curated topic slates and allows for re-ranking the slates based on a user profile provided by RecIt.  

## Todos
- [x] Tests for RecIt request/response code
- [x] Passes tests on re-ranking

## Reference

Tickets:
* [BACK-801](https://getpocket.atlassian.net/browse/BACK-801)

## Implementation Decisions

This is the basic plan, as documented in the ticket:

1.  Pass user_id through in app/models/slate_lineup.py to get_slate_configs_from_experiment
2.  Add exception similar to 'thomson-sampling' in app/models/slate_lineup_config.py:get_slate_configs_from_experiment to make request from slate_ranking branch
3. Make the HTTP request to recit (can be done last while pairing)
4. Add 'curator_topic' metadata to app/json/slate_configs.json
5. Add topic_ranker to app/rankers/algorithms.py that takes in slate list and recit output, and does the mapping


tests/unit/rankers/test_algorithms.py
```diff
@@ -232,3 +236,46 @@ def test_rank_by_ctr_over_n_trials(self, ntrials=99):
...
+
+
+class TestAlgorithmsPersonalizeTopics(unittest.TestCase):
+
+    @staticmethod
+    def _read_json_asset(filename: str):
+        with open(os.path.join(ROOT_DIR, 'tests/assets/json/', filename)) as f:
+            return json.load(f)
+
+    async def test_partial_recit_response(self):
```
mmiermans, line 248:
```suggestion
    async def test_partial_recit_response(self):
        """
        Test that topics are ranked correctly if RecIt returns a subset of
        the topics in the slate lineup config. Topics that are not present
        in RecIt's response should be ranked at the bottom.
        """
```

tests/unit/rankers/test_algorithms.py
```diff
@@ -232,3 +236,46 @@ def test_rank_by_ctr_over_n_trials(self, ntrials=99):
...
+        ordered_output_ids = [c.id for c in output_configs]
+        ordered_output_topics = [c.curator_topic_label for c in output_configs]
+
+        # COVID-19 is not in input_configs
+        # test re-ranking
+        assert ordered_output_topics[:len(personalized_topics)] == personalized_topics
+        # test that all input slates are in output
+        assert set(ordered_input_ids) == set(ordered_output_ids)
+
+    async def test_full_rerank(self):
```
mmiermans, line 267:
```suggestion
    async def test_full_rerank(self):
        """
        Test the case where all topics from the slate lineup config are present in
        the personalized topics list.
        """
```

Pipfile
```diff
@@ -15,6 +15,7 @@ pytest-mock = "*"
 aws_secretsmanager_caching = "*"
 # We define the essential stubs
 boto3-stubs = {extras = ["essential"], version = "==1.16.15.0"}
+aioresponses = "*"
```
mmiermans, line 18: This is similar to the `responses` library, but works with aiohttp.

mmiermans (approved):
:pear: On this. Nice work, @lfishhead!

mmiermans (dismissed):
Thanks for fixing discrepancies between our commits. :bow:

app/models/slate_config.py
```diff
@@ -25,10 +25,11 @@ class SlateConfigModel:
     # store loaded slate configs (loaded at app startup)
     SLATE_CONFIGS_BY_ID = {}
 
-    def __init__(self, slate_id: str, display_name: str, description: str, experiments=None):
+    def __init__(self, slate_id: str, display_name: str, description: str, curator_topic_label=None, experiments=None):
         self.id = slate_id
         self.displayName = display_name
         self.description = description
+        self.curator_topic_label = Optional[curator_topic_label]
```
iwsmith, line 32: The `Optional` modified should go on the `curated_topic_label : Optiona[str]` parameter method, not here.
This doesn't do what you want:
```
>>> from typing import Optional
>>> a = None
>>> b = Optional[a]
>>> type(a)
<class 'NoneType'>
>>> type(b)
<class 'type'>
>>> b
<class 'NoneType'>
>>> a
```

app/rankers/algorithms.py
```diff
@@ -111,6 +112,20 @@ def thompson_sampling(
...
 
 
+async def personalize_topic_slates(input_slate_configs: List['SlateConfigModel'],
+                                   personalized_topics: PersonalizedTopicList) -> List['SlateConfigModel']:
+    """
+    This routine takes a list of slates as input in which must include slates with an associated curator topic
+    label.  It uses the topic_profile that is supplied by RecIt to re-rank the slates according to affinity
+    with items in the user's list.
+    :param input_slate_configs: SlateConfigModel list that includes slates with curatorTopicLabels
+    :param user_id: str indicating user profile for personalization
```
iwsmith, line 122: This param is wrong

app/json/slate_config.schema.json
```diff
@@ -36,6 +36,16 @@
         ],
         "pattern": "^(.+)$"
       },
+      "curatorTopicLabel": {
```
iwsmith, line 39: Consider adding a enum so we don't enter invalid topics: https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

app/rankers/algorithms.py
```diff
@@ -111,6 +112,20 @@ def thompson_sampling(
...
+    :return: SlateLineupExperimentModel with reordered slates
+    """
+    topic_to_score_map = {t.curator_topic_label: t.score for t in personalized_topics.curator_topics}
+    return sorted(input_slate_configs, key=lambda s: topic_to_score_map.get(s.curator_topic_label, 0), reverse=True)
```
iwsmith, line 126: Does this mean non-topic slates all get pushed to the bottom of the lineup? Is there anyway for us to say we only want the top performing topic slate rather than all 15?

iwsmith (changes requested):
I am unclear on how this will function with a lineup that contains topic and non-topic slates. I think the non-topic slates will be shoved to the bottom, which I don't think we want.

app/rankers/algorithms.py
lfishhead: can't we cascade a topN ranker after the personalized ranker?

lfishhead: I am also unclear on how it *should* work in the question you raise, but this seems beyond what recommendations-api currently allows, i.e. a lineup within a lineup.  which seems like a bigger design discussion.  i guess one possibility is to assume a lineup with a subset of topic slates and then a param that determines the number of topic slates that should be preserved after personalization.

app/models/slate_config.py
```diff
@@ -2,16 +2,36 @@
...
+    CAREER = 'Career'
+    EDUCATION = 'Education'
+    ENTERTAINMENT = 'Entertainment'
+    FOOD = 'Food'
+    GAMING = 'Gaming'
+    HEALTH = 'Health & Fitness'
+    PARENTING = 'Parenting'
+    PERSONAL_FINANCE = 'Personal Finance'
+    POLITICS = 'Politics'
+    SCIENCE = 'science'
```
iwsmith, line 23: Should this be upper case?

app/models/slate_config.py
```diff
@@ -25,12 +45,14 @@ class SlateConfigModel:
...
-    def __init__(self, slate_id: str, display_name: str, description: str, curator_topic_label=None, experiments=None):
+    def __init__(self, slate_id: str, display_name: str, description: str,
+                 curator_topic_label: Optional[CuratorTopic] = None, experiments=None):
         self.id = slate_id
         self.displayName = display_name
         self.description = description
-        self.curator_topic_label = Optional[curator_topic_label]
         self.experiments = experiments or []
+        if curator_topic_label:
+            self.curator_topic_label = curator_topic_label
```
iwsmith, line 55: I am not sure what the intent here is, but in this setup `self.curator_topic_label` won't exist as an attribute if `curator_topic_label` is `None`, which means any attempt to access it will raise an AttributeError. If that is desired you should leave a comment as to why; normal behavior is to just assign `self.curator_topic_label = curator_topic_label`, which will default to `None`.

iwsmith (changes requested):
Two minor changes, but they are important.

app/rankers/algorithms.py
```diff
@@ -112,18 +112,27 @@ def thompson_sampling(
     return [x[0] for x in scores]
 
 
-async def personalize_topic_slates(input_slate_configs: List['SlateConfigModel'],
-                                   personalized_topics: PersonalizedTopicList) -> List['SlateConfigModel']:
+def personalize_topic_slates(input_slate_configs: List['SlateConfigModel'],
+                             personalized_topics: PersonalizedTopicList,
+                             topic_limit: Optional[int] = None) -> List['SlateConfigModel']:
```
mmiermans, line 117: This parameter is currently unused, so I suggest we remove it.

It's implemented as a simple limit before returning the result, so it could be replaced by a top_n_ ranker, which I think would be a cleaner solution, because it would keep our rankers focused. I can see how there could be more complex constraints (e.g. remove all but the first X topic slates), but let's cross that bridge when the need is there.
```suggestion
                             ) -> List['SlateConfigModel']:
```

app/rankers/algorithms.py
```diff
@@ -112,18 +112,27 @@ def thompson_sampling(
...
     """
     topic_to_score_map = {t.curator_topic_label: t.score for t in personalized_topics.curator_topics}
-    return sorted(input_slate_configs, key=lambda s: topic_to_score_map.get(s.curator_topic_label, 0), reverse=True)
+    # filter non-topic slates
+    input_slate_configs = list(filter(lambda s: s.curatorTopicLabel in topic_to_score_map, input_slate_configs))
+    if not input_slate_configs:
+        raise ValueError(f"Input lineup to personalize_topic_slates includes no topic slates")
+    # re-rank topic slates
+    input_slate_configs.sort(key=lambda s: topic_to_score_map.get(s.curator_topic_label), reverse=True)
+    return (input_slate_configs[:topic_limit] if topic_limit else input_slate_configs)
```
mmiermans, line 135: Is `topic_limit` limiting all slates here, instead of just topic slates? Let's remove it if it's unused.
```suggestion
    return input_slate_configs
```

app/rankers/algorithms.py
```diff
@@ -112,18 +112,27 @@ def thompson_sampling(
...
     :return: SlateLineupExperimentModel with reordered slates
     """
     topic_to_score_map = {t.curator_topic_label: t.score for t in personalized_topics.curator_topics}
-    return sorted(input_slate_configs, key=lambda s: topic_to_score_map.get(s.curator_topic_label, 0), reverse=True)
+    # filter non-topic slates
+    input_slate_configs = list(filter(lambda s: s.curatorTopicLabel in topic_to_score_map, input_slate_configs))
+    if not input_slate_configs:
+        raise ValueError(f"Input lineup to personalize_topic_slates includes no topic slates")
+    # re-rank topic slates
+    input_slate_configs.sort(key=lambda s: topic_to_score_map.get(s.curator_topic_label), reverse=True)
```
mmiermans, line 134: Let's keep the input unchanged to avoid confusion. Most rankers leave the input unchanged, I believe.

If this is a performance optimization, let's first benchmark it to see what the difference is. For the time being, the number of slates is very small (<< 1000), so I think the difference will be insignificant in relation to the total response time.
```suggestion
    return sorted(input_slate_configs, key=lambda s: topic_to_score_map.get(s.curator_topic_label), reverse=True)
```
(I vaguely recall this coming up on Friday. Sorry if I missed it.)

mmiermans (changes requested):
Happy to pair on these changes. Let's pull it across the finish today. :boat:

app/rankers/algorithms.py
lfishhead: I think this allows for more flexibility in the absence of an actual top N ranker.  right now we have 5, 15, and 30.  I think android already wants top1 and web wants top 3.

app/rankers/algorithms.py
lfishhead: I am preserving input now.

app/rankers/algorithms.py
lfishhead: I am going to leave this for now per the above comments.

app/rankers/algorithms.py
```diff
@@ -127,12 +127,12 @@ def personalize_topic_slates(input_slate_configs: List['SlateConfigModel'],
     """
     topic_to_score_map = {t.curator_topic_label: t.score for t in personalized_topics.curator_topics}
     # filter non-topic slates
-    input_slate_configs = list(filter(lambda s: s.curatorTopicLabel in topic_to_score_map, input_slate_configs))
```
mmiermans: I missed that this variable was assigned to, so the input wasn't affected after all. Thanks for the rename!

mmiermans (approved):
Looks good to me! Thanks for pushing this forward. :bow: :star2:

mmiermans (approved):
I regenerated Pipfile.lock to resolve a conflict by running `rm Pipfile.lock && pipenv lock`.

app/models/slate_config.py
lfishhead: ty.  made the change

app/models/slate_config.py
lfishhead: it should

iwsmith (approved):

PR #271, opened 2021-06-03
Validate expires_at

mmiermans:
# Goal
Require `expires_at` to be present, and be set to an integer timestamp that is well into the future. This is intended to prevent candidate sets from being deleted immediately after they're inserted, if expires_at is accidentally set to a time in the past, which happened on June 1st.

# Reference
Retro from June 1st:
- https://docs.google.com/document/d/1JceLv3p7S9xQbyYbqtrj3cFCg13qotDiDgyMN7XgOvY/edit#

## Implementation Decisions
```python
# expires_at cannot be less that this many seconds from now. It's currently arbitrarily set to 1 day, which is much
# less than the expiration duration used in practice of 1 month. This is intended as a sanity check to ensure that
# expires_at isn't accidentally set to the current time, or X hours from now, instead of X days from now.
MINIMUM_EXPIRES_AT_FROM_NOW = 86400
```


kkyeboah (approved):
nice 👌

PR #273, opened 2021-06-03
Add personalized topics ranker to topic lineup

lfishhead:
# Goal

Enable the ranker that personalizes the order of slates in the topic lineup.

## Reference

Tickets:
* [BACK-801](https://getpocket.atlassian.net/browse/BACK-801)

Follow up from https://github.com/Pocket/recommendation-api/pull/264


mmiermans (dismissed):
I approve of Matt's code.

mmiermans (approved):

PR #276, opened 2021-06-04
print->log so we get source info.

iwsmith:
I noticed a lot of these being logged in our cloudwatch logs, figured I would at least make them a logging statement to make them easier to track down.

Open question is what level issue this is. Some slates I wouldn't be surprised if we don't find clickdata (personalized stuff comes to mind), but for others it indicates an issue (e.g. editors picks).

### DIFF
--- a/app/models/recommendation.py
+++ b/app/models/recommendation.py
@@ -1,6 +1,8 @@
 import aioboto3
 
 from asyncio import gather
+
+import logging
 from aws_xray_sdk.core import xray_recorder
 from boto3.dynamodb.conditions import Key
 from enum import Enum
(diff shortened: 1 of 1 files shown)


app/models/recommendation.py
```diff
@@ -128,6 +130,6 @@ async def __thompson_sample(slate_id: str, recommendations: ['RecommendationMode
             click_data = await ClickdataModel.get(slate_id, item_ids)
         except ValueError:
             rec_item_ids = ','.join(item_ids)
-            print(f'click data not found for candidates with item ids: {rec_item_ids}')
+            logging.warn(f'No click data found for {slate_id = } {rec_item_ids = }')
```
mmiermans, line 133: What does the `= }` do? I couldn't immediately find the meaning in the [f-string spec](https://www.python.org/dev/peps/pep-0498/#specification).
```python
f ' <text> { <expression> <optional !s, !r, or !a> <optional : format specifier> } <text> ... '
```

mmiermans (commented):
Look good, but I'm unfamiliar with the `f{foobar = }` syntax. Glad to keep learning new Python things from you though!

app/models/recommendation.py
iwsmith: It prints out the variable name and its value:
```
>>> a = "corgis are the longest bois"
>>> print(f"{a =}")
a ='corgis are the longest bois'
```

PR #310, opened 2021-07-22
Block link to 400 error code on external page

mmiermans:
# Goal
Block and item that results in a 400 error.

Title: If your resume is missing this number, you can kiss the job away
Url: http://theladders.com:443/career-advice/resume-missing-this-number-goes-in-trash?utm_source=pocket_discover

## Reference

Slack thread:
* https://pocket.slack.com/archives/C2ZEL2NSF/p1626977671004000
* https://pocket.slack.com/archives/C2ZEL2NSF/p1626977114002000

Career page:
* https://getpocket.com/explore/career

Query to get item id by title in [Apollo](https://studio.apollographql.com/graph/pocket-client-api/explorer?variant=current):
```
query Query($getSlateSlateId: String!) {
  getSlate(slateId: $getSlateSlateId) {
    recommendations {
      item {
        itemId
        title
      }
    }
  }
}
```
Variables:
```
{
  "getSlateSlateId": "e6c8db8d-cae7-4947-a57e-872dec639472"
}
```

## Implementation Decisions
- I have not yet looked into whether the parser incorrectly resolved this url in some way. In any case I believe a fix there would only apply after the candidate set is regenerated.


chelseatroy (approved):
I try to avoid "LGTM" but this one is pretty small :)

PR #317, opened 2021-07-28
Create QA slates that are served to a QA user

mmiermans:
# Goal
We want to get some confidence that the engagement pipeline works end-to-end. The approach I'd like to take is to create QA slates for which we can control exactly how many impressions and opens it's gets, because only a particular QA user will see it. This will allow us to track this QA slate using its unique slate id through the engagement pipeline.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HOMPER-10

Documentation:
* https://docs.google.com/document/d/1AHAIdKrgDLPgUgUVoziqnX9vgnFyASBZQ595y3Oqvkk/edit#heading=h.o31clnk4jx7l

QA user details:
* https://pocket.slack.com/archives/C03QVL4SU/p1627576490339900

## Implementation Decisions
- This change is intended to be temporary. Once the end-to-end QA is complete, we can revert this change.
- A better long-term solution would be to use Unleash for deciding which branch of experiments users fall into. Unleash can put particular users into a particular branch. We originally decided not to include Unleash in recommendation-api because we didn't need most of its features. We should reevaluate whether that's still the case, but I'd prefer not to be blocked on that.

### DIFF
--- a/app/models/slate.py
+++ b/app/models/slate.py
@@ -21,6 +21,13 @@ class SlateModel(BaseModel):
     description: str = None
     recommendations: List[RecommendationModel] = None
 
+    __QA_USER_IDS = ['47372502']
+    __QA_SLATE_MAP = {
+        "2e3ddc90-8def-46d7-b85f-da7525c66fb1": "9dc26792-10ed-4fbe-a13d-6cce3a89b0a1",
+        "0c09627b-a409-4768-b87d-7e1d29259785": "e8251442-ef97-422f-ad65-0f28e6f7a0d6",
+        "0f322865-64e6-472d-8147-b3d6637a7d67": "b70d65c6-9171-40bf-bddb-5a60d42dd03f",
+    }
+
     @staticmethod
     @xray_recorder.capture_async('models_slate_get_slate')
     async def get_slate(slate_id: str, user_id: str = None, recommendation_count: Optional[int] = 10) -> 'SlateModel':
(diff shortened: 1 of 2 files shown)


kschelonka (approved):
I suggest ticketing the work to either use Unleash or undo this (if you haven't already done so), but seems fine!

app/models/slate.py
```diff
@@ -21,6 +21,13 @@ class SlateModel(BaseModel):
     description: str = None
     recommendations: List[RecommendationModel] = None
 
+    __QA_USER_IDS = ['47372502']
+    __QA_SLATE_MAP = {
+        "2e3ddc90-8def-46d7-b85f-da7525c66fb1": "9dc26792-10ed-4fbe-a13d-6cce3a89b0a1",
+        "0c09627b-a409-4768-b87d-7e1d29259785": "e8251442-ef97-422f-ad65-0f28e6f7a0d6",
+        "0f322865-64e6-472d-8147-b3d6637a7d67": "b70d65c6-9171-40bf-bddb-5a60d42dd03f",
+    }
```
jpetto, line 29: even though it's temporary code, i think some comments here would be helpful - e.g. commenting the user/email associated with that id, and commenting each slate in the map (most read, personal finance, etc).

jpetto (changes requested):
i think adding comments here is important as it's a shared codebase, but otherwise this looks like a clean & simple way to QA

app/models/slate.py
mmiermans: For future reference, the QA user details are here: https://pocket.slack.com/archives/C03QVL4SU/p1627576490339900

PR #323, opened 2021-07-30
chore(.aws): updating terraform modules

bassrock:
# Goal

Update terraform modules to the latest version and clean up .aws a bit.

## Implementation Decision

I chose to not switch to codepipeline or make other changes.

### DIFF
--- a/.aws/src/config/index.ts
+++ b/.aws/src/config/index.ts
@@ -1,27 +1,23 @@
 const name = 'RecommendationAPI';
-let environment;
-let domain;
-let cacheNodes;
-let cacheSize;
+const domainPrefix = 'recommendation-api';
+const isDev = process.env.NODE_ENV === 'development';
+const environment = isDev ? 'Dev' : 'Prod';
+const isProd : boolean = environment === "Prod";
+const domain = isDev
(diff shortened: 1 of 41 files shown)


mmiermans (approved):

.aws/src/config/index.ts
```diff
@@ -1,27 +1,23 @@
...
+const cacheNodes = isDev ? 1 : 1;
+const cacheSize = isDev ? 'cache.t2.micro' : 'cache.t3.medium';
 
 export const config = {
   name,
   prefix: `${name}-${environment}`,
   circleCIPrefix: `/${name}/CircleCI/${environment}`,
   shortName: 'RECAPI',
+  isProd,
+  isDev,
```
chelseatroy, line 20: This is so minor and I wouldn't block on it, but in general we've been using one or the other of these rather than both. Should we move towards both in other repos?

chelseatroy (approved):
I just made one minor comment but otherwise looks good!

.aws/src/config/index.ts
bassrock: hmm We should have a single one. Let me merge this and then I will followup with a a change in the next one.

PR #329, opened 2021-08-03
Adding a personalized webhome lineup

lfishhead:
# Goal

Add a personalized lineup for the H2 web home work.  The details of the lineups are included [here](https://getpocket.atlassian.net/l/c/WsNaYxD4).  The goal of the work is to allow for lineups to have a mix of topic and non-topic slates and leave the non-topic slates as ordered, and remove the lower ranked topic slates after a cutoff.

## Reference

* [Details](https://getpocket.atlassian.net/l/c/WsNaYxD4)
* [HOMEPER-25](https://getpocket.atlassian.net/browse/HOMPER-25)
* [HOMEPER-23](https://getpocket.atlassian.net/browse/HOMPER-23)
* [HOMEPER-27](https://getpocket.atlassian.net/browse/HOMPER-27)

## Implementation Decisions

We refactored the personalized topic slate ranker to comply with the `topN` style used for truncating candidate sets.  There is now a `top1_topics` ranker used for Android discover and a `top3_topics` ranker for web home.

I remain a little dissatisfied that we include all the topic slates in these lineups which is fine for now but won't scale if the set of topics grows.

### DIFF
--- a/app/rankers/algorithms.py
+++ b/app/rankers/algorithms.py
@@ -62,7 +62,7 @@ def top45(items: RankableListType) -> RankableListType:
 
 def top1_topics(slates: List['SlateConfigModel'], personalized_topics: PersonalizedTopicList) -> List['SlateConfigModel']:
     """
-    returns the lineup with only the top
+    returns the lineup with only the top topic slate included
     :param slates: initial list of slate configs
     :param personalized_topics: recit response including sorted list of personalized topics
     :return: list of slate configs with only the top topic slate
@@ -74,7 +74,7 @@ def top1_topics(slates: List['SlateConfigModel'], personalized_topics: Personali
 def top3_topics(slates: List['SlateConfigModel'], personalized_topics: PersonalizedTopicList) -> List[
     'SlateConfigModel']:
     """
-    returns the lineup with only the top
+    returns the lineup with only the top 3 topic slates included
     :param slates: initial list of slate configs
     :param personalized_topics: recit response including sorted list of personalized topics
     :return: list of slate configs with only the top 3 topic slate
--- a/app/rankers/algorithms.py
+++ b/app/rankers/algorithms.py
@@ -15,7 +15,7 @@ from app.models.personalized_topic_list import PersonalizedTopicList
 DEFAULT_ALPHA_PRIOR = 0.02
 DEFAULT_BETA_PRIOR = 1.0
 
-RankableListType = Union[List['SlateModel'], List['RecommendationModel']]
+RankableListType = Union[List['SlateConfigModel'], List['RecommendationModel']]
 RecommendationListType = List['RecommendationModel']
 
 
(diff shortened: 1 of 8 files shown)


app/rankers/algorithms.py
```diff
@@ -15,7 +15,7 @@
 DEFAULT_ALPHA_PRIOR = 0.02
 DEFAULT_BETA_PRIOR = 1.0
 
-RankableListType = Union[List['SlateModel'], List['RecommendationModel']]
+RankableListType = Union[List['SlateConfigModel'], List['RecommendationModel']]
```
mmiermans, line 18: For future reference: this is an unrelated bug to this PR. We were using the wrong type here, but no ranker using `RankableListType` was used for slates, so this did not have any impact yet.

tests/unit/rankers/test_algorithms.py
```diff
@@ -288,42 +293,50 @@ def test_personalize_topic_limit(self):
...
-            # first config is not personalziable and should be same as input
-            assert input_configs[0] == output_configs[0]
-            # returned topic slate should be highest ranked
-            assert ordered_output_topics[1:1+topic_limit] == personalizable_slates[:topic_limit]
-            # last two output slates are also not personalizable
-            assert input_configs[-2:] == output_configs[-2:]
-
+        personalized_topics = [x.curator_topic_label for x in full_topic_profile.curator_topics]
+
+        for lineup_id in [ANDROID_DISCOVER_LINEUP_ID, WEB_HOME_LINEUP_ID]:
```
mmiermans, line 310: Pytest supports [parameterized tests](https://docs.pytest.org/en/6.2.x/example/parametrize.html), which have the benefit of telling you which parameter failed, and keeping tests a little more readable.

mmiermans (approved):
We :pear:'ed. Looks good to me.

app/rankers/algorithms.py
chelseatroy: Gotcha. Thanks for flagging; I would have been confused :)

app/rankers/algorithms.py
```diff
@@ -60,6 +60,29 @@ def top45(items: RankableListType) -> RankableListType:
...
+    'SlateConfigModel']:
+    """
+    returns the lineup with only the top 3 topic slates included
+    :param slates: initial list of slate configs
+    :param personalized_topics: recit response including sorted list of personalized topics
+    :return: list of slate configs with only the top 3 topic slate
+    """
+
+    return __personalize_topic_slates(slates, personalized_topics, topic_limit=3)
+
```
chelseatroy, line 84: I'm on board with this, especially since we're talking about an API surface of 2 that both point back to one implementation.

If we start having 3 or more `topN_topics` rankers, it might be worth putting our heads together on a more generalizable ranker schema.

app/rankers/algorithms.py
```diff
@@ -131,9 +154,9 @@ def thompson_sampling(
     return [x[0] for x in scores]
 
 
-def personalize_topic_slates(input_slate_configs: List['SlateConfigModel'],
-                             personalized_topics: PersonalizedTopicList,
-                             topic_limit: Optional[int] = 1) -> List['SlateConfigModel']:
+def __personalize_topic_slates(input_slate_configs: List['SlateConfigModel'],
```
chelseatroy, line 157: Might be worth putting in the docstring something to the effect of "Instead of calling this directly, make a new public method that calls it (see `top1_topics` for example)." That way the next person who needs this functionality will be at less risk of defaulting to making this method public again and using it directly.

tests/unit/rankers/test_algorithms.py
```diff
@@ -288,42 +293,50 @@ def test_personalize_topic_limit(self):
                                for t in full_topic_profile.curator_topics
                                if t.curator_topic_label in input_topics]
 
-        for test_limit in [1, 3, 5]:
-            output_configs = personalize_topic_slates(input_configs, full_topic_profile, test_limit)
+        for test_limit, topic_ranker in zip([1, 3], [top1_topics, top3_topics]):
+            output_configs = topic_ranker(input_configs, full_topic_profile)
```
chelseatroy, line 297: I'll enter this as further evidence in support of "if it becomes more than 2, let's rethink how we do this"

tests/unit/utils.py
```diff
@@ -51,11 +50,11 @@ def generate_uncurated_configs() -> List[SlateConfigModel]:
 
     return input_configs
 
-def generate_hybrid_configs() -> List[SlateConfigModel]:
+def generate_lineup_configs(lineup_id: str):
+    # get slate configs from lineup_id
```
chelseatroy, line 54: Maybe make this a docstring

chelseatroy (commented):
Mostly looks good. See comments!

iwsmith (approved):

app/rankers/algorithms.py
lfishhead: +1

PR #333, opened 2021-08-06
Add required_terraform pinning Terraform to 0.13.5

danieldreier:
# Goal

Terraform operations should error out if Terraform is run with newer Terraform versions.

## Todos
- [ ] needs to be run / tested

## Reference
@bassrock, @kkyeboah and I discussed this in slack

I tested this in a toy environment but haven't run the `recommendation-api` CDKTF codebase locally


mmiermans (approved):
lgtm :+1: Thanks for help fix this!

I'll look to @kkyeboah for a second review.

.aws/src/main.ts
```diff
@@ -265,4 +265,5 @@ class RecommendationAPI extends TerraformStack {
 
 const app = new App();
 new RecommendationAPI(app, 'recommendation-api');
+stack.addOverride("terraform.required_version", "0.13.5")
```
kkyeboah, line 268:
```suggestion
const stack = new RecommendationAPI(app, 'recommendation-api');
stack.addOverride("terraform.required_version", "0.13.5")
```

kkyeboah (changes requested):
Thank you for your contribution 💯. We definitely need to require the version.

I do have a small suggestion to fix your addition.

mmiermans:
We've since migrated to Terraform 1.0.11. Not sure if pinning is still necessary.

PR #334, opened 2021-08-06
Propose solution for graphql fallback behavior for personalized slates.

jeshuaborges:
# Goal

Define a standard for fallback behavior when a personalized slate can not be generated.

## Todos
- [x] Document Graphql

## Review
- [ ] GraphqlExperts review @Pocket/backend 

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HOMPER-26

## Implementation Decisions

On the #p-user-profiles project we are adding some new behavior into the RecsAPI. The new behavior will be around querying personalized slates. What we want to do is to allow clients to request a personalized slate lineup and in the scenario where there is no user profile, for generating a slate lineup, we will return a nonpersonalized slate lineup as a fallback.  Where the simplest is to swap the slate lineup id on the response. However this would be an unexpected behavior as when a client requests a specific piece of content, we should return an error if that item doesn't exist. To satisfy this new requirement I am thinking that we should add something more explicit.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -63,6 +63,7 @@ type Query {
 
     query getPersonalizedSlateLineup($id: String!) {
         getPersonalizedSlateLineup(id: $id) {
+            __typename
             personalizedSlateLineup {
                 ...SlateLineupParts
             }
--- a/schema.graphql
+++ b/schema.graphql
@@ -297,9 +297,14 @@ A requirement for personalization is missing. Use this slate linup instead.
 """
 type PersonalizedSlateLineupFallback {
     """
+    Why the fallback was used.
+    """
+    error: String!
+    
+    """
     The fallback slate lineup to be used when the personalized slate lineup is not available.
     """
-    fallbackSlateLineup: SlateLineup
+    fallbackSlateLineup: SlateLineup!
 }
 
 extend type Item @key(fields: "itemId") {
--- a/schema.graphql
+++ b/schema.graphql
@@ -38,6 +38,48 @@ type Query {
         "Maximum number of recommendations to return in {Slate.recommendations}, defaults to 10"
         recommendationCount: Int = 10
     ): SlateLineup
+
+
+    """
+    Request a specific personalized `SlateLineup` by id. This response allows for fallback behavior when there is not
+    a personalized slate lineup available.
+
+    Examples:
+
+    fragment SlateLineupParts on SlateLineup {
+        id
+        experimentId
+        slates {
+            recommendations {
+                id
+                item: {
+                    resolvedUrl
+                    title
+                    topImageUrl
+                }
+            }
+        }
+    }
+
+    query getPersonalizedSlateLineup($id: String!) {
+        getPersonalizedSlateLineup(id: $id) {
+            personalizedSlateLineup {
+                ...SlateLineupParts
+            }
+            fallbackSlateLineup {
+                ...SlateLineupParts
+            }
+        }
+    }
+    """
+    getPersonalizedSlateLineup(
+        "The {SlateLineup.id} of the SlateLineup to return"
+        slateLineupId: String!,
+        "Maximum number of slates to return in {SlateLineup.slates}, defaults to 8"
+        slateCount: Int = 8,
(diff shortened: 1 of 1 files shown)


kschelonka:
I like how explicit this is. Would the intent be that the response would include both `data` and `error` fields if the personalized option didn't exist? Or just that the personalized option would be null (and error would remain undefined)?

kkyeboah:
@jeshuaborges this is a good solution for a fallback to me. @kschelonka has a good question here about how we plan to structure the response. I have thoughts but I'd like to hear your answer to the question.

schema.graphql
```diff
@@ -38,6 +38,48 @@ type Query {
...
+    query getPersonalizedSlateLineup($id: String!) {
+        getPersonalizedSlateLineup(id: $id) {
+            personalizedSlateLineup {
+                ...SlateLineupParts
+            }
+            fallbackSlateLineup {
+                ...SlateLineupParts
+            }
+        }
+    }
```
chelseatroy, line 73: I'm into this. I have seen a similar pattern in a number of API responses wherein the JSON always included `data` and `error`, and if there was no `data`, the client would get the `error` field to figure out what went wrong.

This is also clearer than an in-situ replacement without a message, and will result in no more code on the client than checking that the returned lineup's ID matched the requested one would. 

The fact that we send the fallback every time also alleviates a concern about making a second request in the majority of cases (since, as I understand it, unpersonalized behavior will still be our majority case for now).

Finally, the potential drawback I could anticipate with this request (and it's not guaranteed) is payload size. For now I'd actually say ignore that until and unless it becomes an issue. If it _does_, we could discuss options like only populating the fallback if the `personalizedSlateLineup` is not available.

chelseatroy: @iwsmith @lfishhead @mmiermans thoughts?

jeshuaborges:
@kschelonka 

> Would the intent be that the response would include both data and error content if the personalized option didn't exist? 

The intent is that the query will return a `PersonalizedSlateLineup`. This means that it would return either `PersonalizedSlateLineupSuccess` or `PersonalizedSlateLineupFallback`. The client can query the `__typename` parameter to know which type of response it received. Additionally, I just added an `error` property to the `PersonalizedSlateLineupFallback` so that we could more explicitly give a reason to the client.

My logic may be wrong in how Graphql union types work and may schema may not work the way I am intending to document.

> Or just that the personalized option would be null (and error would remain undefined), but the client would know a 'not-found' error occurred since it's null?

With the `PersonalizedSlateLineupFallback`, no personalized recommendations, the `personalizedSlateLineup` property will be null and the `fallbackSlateLineup` and `error` keys would be populated.

schema.graphql
iwsmith: Personally I don't like this solution. Sending both the personalized and unpersonalized responses seems ripe for confusion from clients and couples clients to the implementation details of what kind of recommendations are being generated. Furthermore, what will the clients do with this information (i.e. that the personalized request failed)? They can't fix this problem, there is no action they or their users can take. In practice they are just going to check for one field, and if it is missing use the other field instead. So I don't see why we wouldn't save them the trouble and potential confusion and do it server side.

As for "when a client requests a specific piece of content, we should return an error if that item doesn't exist", there is a philosophical question about what a SlateLineup is. Originally, I created them with the intent of identifying a specific place in a specific client where content is displayed. In this respect personalization doesn't matter, that is an implementation detail for RecsAPI. Similarly, the order the slates are returned in is an implementation detail. All the clients care about is that suitable content for this location is returned. We as RecsAPI care, but the clients shouldn't. 

Note that there has been a proliferation of SlateLineups, so that many different ones can now represent that same client/location. I think that is b/c we ended up using new SlateLineups to run experiments rather than setting them up as an experiment inside the SlateLineup (I suspect this is b/c we never implemented that functionality in RecsAPI).

iwsmith: Ah reading below I see that the intent here is to use a typed union. If so just ignore most of what I wrote above.

lfishhead:
I don't have more to add other than the clients appear to then have the responsibility to identify the correct lineup ID as well as whether they are requesting a personalized lineup or not.  I think that's ok, but am interested in others' thoughts.

PR #335, opened 2021-08-06
slatelineup for firefox android home recommendations

jpetto:
# Goal

create a slatelineup for firefox android home recommendations. 

PRD:
* https://docs.google.com/document/d/1-P-Q15nHMGy_TQuwFh3YW07-dINXB5_rDMocwTuFl5s/edit#heading=h.jyahmgu05z6p (direct link to slates requested for this lineup)

### DIFF
--- a/app/json/slate_lineup_configs.json
+++ b/app/json/slate_lineup_configs.json
@@ -513,5 +513,27 @@
         ]
       }
     ]
+  },
+  {
+    "id": "a26db39b-a68a-4e01-b680-443a466f9c36",
+    "description": "Firefox Android Home Recommendations",
+    "experiments": [
+      {
+        "description": "default layout",
+        "rankers": [],
+        "slates": [
+          "0737b00e-a21e-4875-a4c7-3e14926d4acf",
+          "2e3ddc90-8def-46d7-b85f-da7525c66fb1",
+          "626397d3-fa4e-4299-8d07-cbeaa269ebc1",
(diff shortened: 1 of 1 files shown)


jpetto:
@mmiermans i don't think i changed anything that would cause this test to fail? do you have any quick insight here?

mmiermans (dismissed):
Sorry for the flaky test. Fix for that is here: https://github.com/Pocket/recommendation-api/pull/338

app/json/slate_lineup_configs.json
```diff
@@ -513,5 +513,27 @@
         ]
       }
     ]
+  },
+  {
+    "id": "a26db39b-a68a-4e01-b680-443a466f9c36",
```
jpetto, line 518: i just generated this in the python console - is that okay?

chelseatroy: Yes. At some point we discussed these IDs being a hash of the object, but then we truncated it, at which point it doesn't represent the object anymore. Random seems fine.

chelseatroy (approved):
Slivin'

app/json/slate_lineup_configs.json
iwsmith: Just to share knowledge: the slate/lineup ids are static GUIDs, the only requirement is that they are unique w.r.t the namespace. The experiment_ids are truncated hashes based on the slate/lineup configuration, and are auto-generated.

kkyeboah (approved):
lgtm

PR #338, opened 2021-08-09
Fix assertion that would fail randomly

mmiermans:
# Goal
Fix test that would fail 50% of time time, because the order of recommendations was determined using Thompson sampling.

## Review
Any ideas to prevent similar mistakes from happening in the future, beyond the comments that I added?

## Reference

CircleCI:
* https://app.circleci.com/pipelines/github/Pocket/recommendation-api/1511/workflows/a79d0370-c564-4b3d-8617-89a2a844f829/jobs/5841


jpetto (approved):

jpetto:
not sure how much work it would be, but only other option i can see is removing thompson sampling from the rankers for these tests. however, as long as order doesn't matter for these tests, the fix you've implemented seems fine.

PR #340, opened 2021-08-10
replace travel with career slate in firefox android home lineup

jpetto:
# Goal

the curated career slate is more reliably populated than the curated travel slate, so we're swapping out travel for career.

career slate is documented here: https://getpocket.atlassian.net/wiki/spaces/PE/pages/2046197833/RecsAPI+Slate+Documentation#Curated-Career-Slate


kschelonka (approved):
💁‍♀️

PR #343, opened 2021-08-11
feat(slate meta data): separate UI content

collectedmind:
# Goal
Create easily renderable meta-data without needing to know what is present in the lineup.

## Todos
- [x] Update Schema
- [x] Update json configs
- [ ] Connect the dots (unfamiliar with the setup here so I am sure I missed several things)

## Review
- [x] General @lfishhead 

## Implementation Decisions

In order for the front end to consistently render titles without fear of exposing internal naming (`Coronavirus Slate` vs `Coronavirus`) or getting conjoined returns of title/subtitle this just proposes a couple of new fields to add to the metadata.

1. `displayTitle`: This is the title to use for a given slate if required
2. `displaySubTitle`: This is the subtitle to use if required
3. `displayTopic`: This is the topic that will be used if required/available

### DIFF
--- a/app/json/slate_configs.json
+++ b/app/json/slate_configs.json
@@ -2,6 +2,7 @@
   {
     "id": "de254c5d-57a7-4553-850f-153ee385014d",
     "displayName": "Editors' Picks",
+    "displayTitle": "Editors' Picks",
     "description": "Curated items that are over a week old",
     "experiments": [
       {
@@ -31,6 +32,7 @@
   {
     "id": "2389f563-bd42-411d-bca2-0966638053e8",
     "displayName": "Editors' Picks",
+    "displayTitle": "Editors' Picks",
     "description": "Curated items that are over a week old without syndicated",
     "experiments": [
       {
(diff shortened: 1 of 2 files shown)


lfishhead:
I see this as a stopgap in the sense that we all wish for more and better copy associated with the slates and lineups, and even the ability to experiment with the copy itself.  As this repo is configured, any change to this information will require a PR and back end approval.  Copy isn't code.  We should design a better means for marketing or editorial to be able to make changes here that don't involve engineering.  Thanks to @collectedmind  for trying to move things forward in the meanwhile.

collectedmind:
> Copy isn't code

Agree wholeheartedly.  In the meantime, as we move to production with some of these slates, it will be a challenge to achieve the fluidity we want with lineups and slates without the stopgap.

collectedmind:
I am thinking I would also like to ad a `displayType` but I am still working through the idea there and it is a bit of a rabbit hole

jeshuaborges:
@lfishhead How would you feel about closing this PR?

mmiermans:
The challenge of managing slate copy still unsolved. We are tracking this as a project in our backlog: https://getpocket.atlassian.net/wiki/spaces/CP/pages/2509897803/Draft+Data+Products+Meta+Data

PR #347, opened 2021-08-13
Fetch the userId from the query itself rather than a header

chelseatroy:
# Context

It looks like we expect personalized slate lineup requests to include the user_id as a header that we resolve at the middleware level, rather than as part of the GQL query itself.

## Question
How did we decide on that approach? 

## Proposal

I'm picturing an alternative in which the client includes the userId in the query as a parameter—just like slateLineupId, except optional. This is more graph queuelly, but the functional advantages would be:

- One centralized location for visualizing the passage of information about the expected result to the Recs API
- We can use Apollo client to make personalized queries, rather than having to shell out to Postman and use a RESTified version of the query to run it, which is what we are doing now.

This PR makes the change, so it wouldn't create extra work for anyone to implement. But before I do, I'd love to know:

1. If I've overlooked a requirement that influenced us to do it the way it's currently done
2. How many places we'll have to change clients if we change this API

### DIFF
--- a/app/graphql/graphql.py
+++ b/app/graphql/graphql.py
@@ -26,8 +26,9 @@ class Query(ObjectType):
                                                description="Number of recommendations to return, defaults to 10"))
     list_slates = Field(List(Slate), recommendation_count=Int(default_value=0,
                                                               description="Number of recommendations to return, defaults to 0"))
-    get_slate_lineup = Field(SlateLineup, slate_lineup_id=String(required=True,
-                                                                 description="SlateLineup id to get a specific slate lineup"),
+    get_slate_lineup = Field(SlateLineup,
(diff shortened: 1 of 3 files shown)


mmiermans (changes requested):
> How did we decide on that approach?
- I'm not sure what the answer is. @bassrock or @kkyeboah might know. My very limited understanding is that client-api extracts the user id from a token that it's able to verify for authenticity.
- If there's one thing I would change with our approach here is to use masked user ids, which can't be iterated over.
- I don't think this is the right approach though. What would prevent a user from impersonating someone else, to see what recommendations that user would get? Personalized recommendations should be treated as personal data.

jeshuaborges:
This is not the approach that we would want to take. @mmiermans' comment is correct.

PR #349, opened 2021-08-17
Fetch more items from RecIt, make top-n rankers warn about insufficient items

iwsmith:
The warning might end up being way too noisy, so I can be convinced to turn that down to an info or debug, but we've had several times now where we intended to get more items from a CandidateSet than we did, and it has caused issues. 

Long term, we probably need an entire configuration to represent how many items are expected at each stage. I did this in RecIt by having rankers explicitly note the min items they expect. We will probably need to have CandidateSets support limits to allow for us to express the idea that we want a specific number of items at each stage of the pipeline.


lfishhead (approved):
the use of partial is so much cleaner.  

as for the logging, I think this is currently an issue that people are attending to, so leaving it as is for now makes sense to me.

jpetto (approved):
lgtm - and oooh `partial` is fancy.

mmiermans (approved):
I keep learning new things from you, Ian! :star2:

PR #372, opened 2021-09-13
Android Personalization V2

iwsmith:
This enables three new slates in support of an Android PersonalizationV2 experiment:
*control* - We are using a different Lineup ID so I can easily differentiate users in experiment from those not in experiment

*personalized_curated* - A new personalization algorithm that generates more diverse recommendations, only recommends items that went live in the last 52 weeks, and generates more items to decrease the amount of repeats users see.

*personalized_topic* - Selects a single topic to show the user based on their profile, but this time shows them an algorithmic topic slate instead of a curated one.

See https://docs.google.com/document/d/1PSjmTiBWIn_tuweoc5kaAVlsEqgv29ujMFoHVUaU0is/edit#

### DIFF
--- a/app/json/slate_lineup_configs.json
+++ b/app/json/slate_lineup_configs.json
@@ -67,6 +67,23 @@
     ]
   },
   {
+    "id": "e39bc22a-6b70-4ed2-8247-4b3f1a516bd1",
+    "description": "iOS Discover Tab",
+    "experiments": [
+      {
+        "description": "Initial layout",
+        "rankers": [],
+        "slates": [
+          "2e3ddc90-8def-46d7-b85f-da7525c66fb1",
+          "0f322865-64e6-472d-8147-b3d6637a7d67",
+          "626397d3-fa4e-4299-8d07-cbeaa269ebc1",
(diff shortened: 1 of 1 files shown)


jpetto (dismissed):
only concern is the `id` change, but i trust you've confirmed there are no clients relying on the existing `id`.

app/json/slate_lineup_configs.json
```diff
@@ -1,7 +1,7 @@
 [
   {
-    "id": "20481076-2261-4916-801a-fd186986367b",
-    "description": "Android Discover Tab Personalized Curated",
+    "id": "aaf47c0f-1361-4c8c-a89f-fa45e1dc2978",
```
kkyeboah, line 3: same concern as @jpetto here

kkyeboah (approved):

app/json/slate_lineup_configs.json
iwsmith: That ID was used in an old experiment, but just to be safe I re-added the old lineup id. It shouldn't be getting used but I am still seeing some analytics events for it 🙃

kkyeboah (approved):

PR #373, opened 2021-09-15
Updating slate descriptions

lfishhead:
# Goal

Carolyn and @collectedmind have provided new text to describe various slates for web home.  These have been added to slate_configs.json to update the `description` field (which is a required field).  

Previous descriptions of the properties of the slates' items have been moved to an `internal_description` field to add in understanding characteristics of the items the slates contain if viewing the config json.  

## Reference

[slate names doc](https://docs.google.com/document/d/1Ha6DbbenMrBOm42uVpH7chRqCrpM-MSwwowulCG-mkI/edit?usp=sharing)

## Implementation Decisions

Moved old descriptions to new non-required `internal_description` field

### DIFF
--- a/app/json/slate_config.schema.json
+++ b/app/json/slate_config.schema.json
@@ -70,7 +70,18 @@
         "title": "The Description Schema",
         "default": "",
         "examples": [
-          "Curated articles within the last week that have high cascade probability"
+          "New perspectives, intriguing deep-dives, and timeless classics"
+        ],
+        "pattern": "^(.+)$"
+      },
+      "internalDescription": {
+        "$id": "#/properties/internalDescription",
+        "type": "string",
+        "title": "The Internal Description Schema",
+        "description": "A slate description for internal use only",
+        "default": "",
+        "examples": [
+          "Curated items including syndicated that are at most a week old"
         ],
         "pattern": "^(.+)$"
       },
--- a/app/json/slate_configs.json
+++ b/app/json/slate_configs.json
@@ -1,8 +1,8 @@
 [
   {
     "id": "de254c5d-57a7-4553-850f-153ee385014d",
-    "displayName": "Editors' Picks",
-    "description": "Curated items that are over a week old",
+    "displayName": "Must-read stories",
+    "description": "Compelling articles worth your time",
     "experiments": [
       {
         "description": "default",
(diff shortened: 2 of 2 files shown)


collectedmind:
giphy

Awesome!

jpetto:
i love this update - only question is, do we know all downstream/client effects are accounted for? have all owners of all systems impacted by this change been notified? (i do not know how to determine this list of owners. 😢 )

lfishhead:
This is a good point.  I am hoping @collectedmind can represent the web team in reviewing this.  @marcin-kozinski can you tell us if android is using the `description` field such that these changes pose any problems?  Hopefully the changes from editorial can benefit all the clients.

marcin-kozinski:
@lfishhead We don't use `description`, we only use `displayName`.

I don't know if all of these slates are eligible to show on Android Discover Tab, so maybe this isn't a problem, but we don't handle long `displayName`s particularly well. Don't get me wrong, I love this copy: "Incredible long reads that’ll transport you to another time and place", but unfortunately this is how it looks on an Android phone (we cut it off):
Long slate display name gets truncated

I don't think this is a huge deal and I'm fine with merging this change. But please let me know if this slate can be shown on Android Discover Tab? If yes, I will speak with Laurie about maybe ticketing some work and getting some design time to look at how we could handle longer titles.

lfishhead:
is the right thing to do here to add a mobile description field also?  I am fine with a separate PR for that after editorial or another team can weigh in on the appropriate copy there.

jpetto:
it feels like relying on a specific textual description downstream is inherently brittle - though i don't know if we can avoid it given the current structure. does each implementing team need their own description field in a slate? (this is only a half-serious question.)

marcin-kozinski:
I think we can use the current `displayName` field, keep slate names consistent across platforms and update the mobile design to accommodate longer titles. Unless Editorial or another team would actually like having separate display names for mobile.

wogz:
+1 to Marcin's recommendation of updating the mobile design and keeping slate names consistent for now. Can we get this code reviewed today? I'd like to get these names updated in time for the Home presentation at the MoCo meeting on Wednesday; slides are due at 12 pm ET tomorrow.

mmiermans (commented):
Overall looks good. Two improvements I'd like to make:
1. Update the JSON schema, primarily to describe that the internal description does.
2. Make the case consistent. I think we use camelcase in most places for JSON.

I implemented above changes here: https://github.com/Pocket/recommendation-api/pull/381

mmiermans (approved):
:+1:

PR #375, opened 2021-09-16
Add topics to algorithmic slates, and topics to slate schema definition

iwsmith:


app/json/slate_configs.json
```diff
@@ -221,6 +221,7 @@
     "id": "d9c6ddfa-a58d-402d-9664-4190ba197ea6",
     "displayName": "Popular with Pocket readers  Stories from across the web",
     "description": "Algorithmic Business Slate",
+    "curatorTopicLabel": "Business",
```
lfishhead, line 224: the reason I didn't have this to start was that the curator labels are manual for the curated slates, but these are all estimates.  it's possible that we might want to make that distinction explicit with something like `estimatedTopicLabel` for the field, but I don't feel strongly.

lfishhead (approved):

app/json/slate_configs.json
iwsmith: Understandable. Functionally we are using it to just match slates to predicted topic labels for a user vector. I think at this point the distinction is more that `curatorTopicLabels` are a specific ontology rather than saying curators choose these labels, so I am ok with keeping the label the same. I could see us having `userTagTopicLabel` in the future which is the ontology based on user tags.

bassrock (approved):
i am assuming we tested locally and new config options don't break any code paths.

This is a great change!

PR #379, opened 2021-09-17
Switch to codepipeline

bassrock:
# Goal

All the rest of our repos are now using codepipeline with dep bot auto merging since codepipeline queues up deploys

This moves this repo to that pattern.


mmiermans (approved):
Yay! :raised_hands:

PR #388, opened 2021-09-21
Change personalized curated text to not duplicate phrasing from other slates

iwsmith:
See https://pocket.slack.com/archives/C0272V883RD/p1632252292160500

Also updated algorithmic topic slate display names so they show up well on Android and are meaningful.


mmiermans (approved):

iwsmith:
I've confirmed that the algorithmic topic slates are only currently used on explore topic pages and joel has confirmed that the display name for the algorithmic slates isn't used to populate any text on that page.

mmiermans (approved):

PR #398, opened 2021-09-27
Add rank-topics ranker, change Web to use it, add all topics

iwsmith:
1. Add `rank-topics`, which just re-ranks the topics based on a user profile
2. Updated the Web SlateLineup to use it for the Web Home Personalization experiment
3. Update the Web Home Personalized and Web Home Unpersonalized SlateLineups to return *all* topics. This will allow for filtering to relevant topics on the client. This may cause performance issue, but since the topics are all fairly static the hope is everything is sufficiently cached.

### DIFF
--- a/app/json/slate_lineup_config.schema.json
+++ b/app/json/slate_lineup_config.schema.json
@@ -92,7 +92,8 @@
                   "top30",
                   "thompson-sampling",
                   "top1-topics",
-                  "top3-topics"
+                  "top3-topics",
+                  "rank-topics"
                 ],
                 "title": "The Items Schema",
                 "default": "",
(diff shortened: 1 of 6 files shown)


lfishhead (approved):
looks great

mmiermans (approved):
:+1: lgtm

I confirmed the latency alarm is set to non-critical, so nobody will be woken up if latency spikes because of this. I'm assuming this means we're ok with evaluating performance in production, as opposed to load testing this change?

The tests are still readable to me, but it would be nice if at some point in the future data engineering would find a better pattern for running the same test with different parameters. I don't think that should block this PR.

PR #403, opened 2021-09-29
Register personalized ranker in a second location, remove Thompson-Sampling from collections

iwsmith:
I missed a location for adding a personalized ranker (see the below sentry error). I centralized this in `rankers/__init__.py` so this doesn't happen again.
https://sentry.io/organizations/pocket/issues/2680882333/?project=5503693&query=is%3Aunresolved

Also remove Thomspson-Sampling from collections, which is unrelated but is necessary to fix collections we think.


mmiermans (approved):
:+1:

lfishhead (approved):

PR #404, opened 2021-09-29
adding new thomspon sampling rankers

lfishhead:
# Goal

The clickdata available in dynamo for Thompson sampling includes rollups of opens and impressions for 1, 7, 14, and 28 day time windows.  This PR adds rankers that can use each of the different time windows that are based on the existing Thompson sampling code previously deployed with a default time window of 28 days.

This can be a part of the solution for making slates that rely on engagement data covering more static collections of items (i.e. syndicated items) more dynamic.

## Reference

[slack](https://pocket.slack.com/archives/C01RJTJBLQ6/p1632871782023400)

## Implementation Decisions

The syndicated slate that is in the web home test has been updated to use the 1-day thompson sampling data which will hopefully be more dynamic.  Additional changes in the construction of the corresponding candidate set for this slate will also be required in the metaflow repo.


app/json/slate_config.schema.json
```diff
@@ -145,6 +145,9 @@
                   "top15",
                   "top30",
                   "top45",
+                  "thompson-sampling-1day",
+                  "thompson-sampling-7day",
+                  "thompson-sampling-14day",
                   "thompson-sampling"
```
iwsmith, line 151: I'd like this change to `thompson-sampling-28day` rather than being implicit

iwsmith (approved):
I'd like to see `thompson-sampling` be explicit about the time window it uses rather than implicit. It will also be more consistent with the other TS rankers. Otherwise seems good.

app/json/slate_config.schema.json
lfishhead: this change is only failing 79 of the 95 tests so far

iwsmith (approved):

app/models/recommendation.py
```diff
@@ -98,16 +98,17 @@ async def get_recommendations_from_experiment(
 
         # apply rankers from the slate experiment on the candidate set's candidates
         for ranker in experiment.rankers:
-            if ranker == 'thompson-sampling':
+            if ranker in THOMPSON_SAMPLING_RANKERS:
```
mmiermans, line 101: Mentioned this to Matt already, but I'd love to spend a fun Friday to see if we can configure the metadata directly on the ranker using a function decorator. That would save us from having to configure that in the __init__.py, and we wouldn't need to make changes when we add a simple `partial` function. :crossed_fingers:

app/models/recommendation.py
```diff
@@ -98,16 +98,17 @@ async def get_recommendations_from_experiment(
...
                 # thompson sampling takes two specific arguments so it needs to be handled differently
-                recommendations = await RecommendationModel.__thompson_sample(slate_id, recommendations)
+                recommendations = await RecommendationModel.__thompson_sample(ranker, slate_id, recommendations)
                 continue
             recommendations = get_ranker(ranker)(recommendations)
```
mmiermans, line 105: Doesn't have to be in this PR, but I'd like to apply the same pattern here that's used for ranking slates:
```
if ranker in THOMPSON_SAMPLING_RANKERS:
    ranker_kwargs = {
        'metrics': await SlateMetricsFactory(dynamodb_endpoint=dynamodb["endpoint_url"]).get(
            slate_lineup_id,
            [s.id for s in slate_configs])
    }

slate_configs = get_ranker(ranker)(slate_configs, **ranker_kwargs)
```
I think that simplifies the code because there's only a single call to the ranker.

app/models/recommendation.py
```diff
@@ -98,16 +98,17 @@ async def get_recommendations_from_experiment(
...
 
         return recommendations
 
     @staticmethod
-    async def __thompson_sample(slate_id: str, recommendations: ['RecommendationModel']) -> ['RecommendationModel']:
+    async def __thompson_sample(ranker: str, slate_id: str,
```
mmiermans, line 110: Nit: I like to clearly distinguish an id/name from the object that does the actual work.
```suggestion
    async def __thompson_sample(ranker_name: str, slate_id: str,
```

app/rankers/__init__.py
```diff
@@ -22,7 +25,10 @@ def get_all_rankers():
         top15,
         top30,
         top45,
-        thompson_sampling,
+        thompson_sampling,  # this uses a 28 day window by default
```
mmiermans, line 28: Nit: I would've just added a `thompson_sampling_28day` partial function to make this explicit all the way.

app/rankers/algorithms.py
```diff
@@ -130,9 +138,9 @@ def thompson_sampling(
         d = metrics.get(clickdata_id)
         if d:
             # TODO: Decide how many days we want to look back.
-            clicks = max(d.trailing_28_day_opens + alpha_prior, 1e-18)
+            clicks = max(getattr(d, opens_column) + alpha_prior, 1e-18)
             # posterior combines click data with prior (also a beta distribution)
-            no_clicks = max(d.trailing_28_day_impressions - d.trailing_28_day_opens + beta_prior, 1e-18)
+            no_clicks = max(getattr(d, imprs_column) - getattr(d, opens_column) + beta_prior, 1e-18)
```
mmiermans, line 143: Nits:
- I'd put the `getattr` calls in variables for readability.
- These variables should probably refer to 'opens' instead of 'clicks'.
- I try to avoid abbreviations like imprs because new developers might not know what it means, even if it seems obvious to us. For example, people have to be told that "recs" means "recommendations", even though in my mind they're totally synonymous by now.

Just a suggestion. Please check for accuracy:
```suggestion
            impression_count = getattr(d, impressions_column)
            open_count = getattr(d, opens_column)
            # Compute beta distribution parameters from open and impression counts.
            # Priors provide a boost to explore items, but impression and open counts will dominate the equation as we get more data.
            opens_parameter = max(open_count + alpha_prior, 1e-18)
            no_opens_parameter = max(impression_count - open_count + beta_prior, 1e-18)
```

mmiermans (approved):
Nice work! :+1: I only had some minor suggestions, so up to you whether you implement those in this PR or not.

PR #406, opened 2021-09-30
fix(apollo): fix apollo service check circleci job

bassrock:
# Goal

I noticed that the copy pasta I did a while back was incorrect and missed a rename, this fixes that.


mmiermans (approved):
Thanks for fixing this! :muscle:

Starting a Python template is high on my fun Friday list, so we can easily replicate all the good stuff in this repo.

PR #408, opened 2021-09-30
Web Home Personalized top3 topics, only return 3 topics for unpersonalized

iwsmith:
Basically resets to what we had a few days ago. Pulling all the topic slates is too slow and we aren't able to address that right now.

### DIFF
--- a/app/json/slate_lineup_configs.json
+++ b/app/json/slate_lineup_configs.json
@@ -554,7 +554,7 @@
     "experiments": [
       {
         "description": "default layout",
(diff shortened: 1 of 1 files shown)


mmiermans (approved):
:+1:

We confirmed the slow-down is caused by recommendation API, and we can get more visibility into Recommendation API by fixing the X-Ray daemon, which is currently failing to start in ECS.

More context: https://pocket.slack.com/archives/C01CKCB9746/p1633024111029200?thread_ts=1632950440.020200&cid=C01CKCB9746

PR #409, opened 2021-10-01
Remove getTopicRecommendations query

mmiermans:
# Goal
Delete code for the getTopicRecommendations query, to simplify this code base and prevent [false alarms in #log-backend-alarms](https://pocket.slack.com/archives/C016NAMSW2J/p1632993124001000). This query is believed to be unused. See 'todos' below for details.

## Todos
- [x] Search for `getTopicRecommendations` in our code base: https://github.com/search?q=org%3APocket+getTopicRecommendations&type=code
    - [x] Confirm that the reference in pocket-ios is unused: https://github.com/Pocket/pocket-ios/blob/main/PocketKit/Sources/Sync/schema.json#L3189
    - [x] Confirm that the reference in curation-admin-tools is unused: https://github.com/Pocket/curation-admin-tools/blob/main/src/collections/api/client-api/generatedTypes.ts#L507

## Review
- Would it be better to leave a dummy `getTopicRecommendations` in place until the above references have been removed? Or should we just confirm that they're not actually used in production (which I'm fairly sure they're not).

## Reference

Slack threads:
- [Confirming iOS does not use this](https://pocket.slack.com/archives/C03C9QQE0/p1633120799133400)
- [Confirming that curation-admin-tools does not use this](https://pocket.slack.com/archives/C03QVL4SU/p1633121814288900)

Follow-up work:
- Remove from [load test](https://github.com/Pocket/load-testing/blob/main/explore-topics/script.yml#L55-L84)
- Remove from [iOS schema](https://github.com/Pocket/pocket-ios/blob/main/PocketKit/Sources/Sync/schema.json#L3188-L3240)
- Remove from [curation-admin-tools schema](https://github.com/Pocket/curation-admin-tools/blob/main/src/collections/api/client-api/generatedTypes.ts#L503-L507) by regenerating that schema in Apollo

## Implementation Decisions
- Keep the Dynamodb candidates table around for a week so we can easily revert this change if needed.
- Keep the `listTopics` query around for now because it's [used in Metaflow](https://github.com/Pocket/dl-metaflow-jobs/blob/main/jobs/recs_api.py#L27-L39).


gkatre (approved):
Looks good. @mmiermans 

>Would it be better to leave a dummy getTopicRecommendations in place until the above references have been removed? Or should we just confirm that they're not actually used in production (which I'm fairly sure they're not).

Is there a way to be really sure that they're not actually used in production? Otherwise may be better make sure things do not break if `getTopicRecommendations` gets referenced.

PR #422, opened 2021-10-13
improve pacing in algorithmic topic slates: first attempt 

lfishhead:
# Goal

Recent android personalization experiment analysis indicates that item churn on the algorithmic topic slates is poor.  One mitigation strategy is to reduce the aggregation window used in Thompson sampling from the default of 28 days.  This change deploys an experiment on those slates that compares windows of 1 day, 7 days, and 28 days (for control).  

This is a first step along building up more levers of control over pacing that will include impression capping and revisions to the item candidate generation query.

## Implementation decisions

After the previous experiment showed a lift in using a 45 candidates for Thompson sampling compared to 30, we use the `top45` ranker in all cases here.

## References

[previous experiment](https://app.mode.com/getpocket/reports/4c428d674696/runs/097e7f50df74) on the algorithmic topic slates


lfishhead:
this makes sense.  i want to get it out so I am not going to change it now, but will try to remember to update it next time I am here.  i am also up for pairing on that change at some point.

PR #439, opened 2021-10-22
API test for getSlateLineup

chelseatroy:
## Context

Prior to this, if we commented out the `get_slate_lineup` resolver in `graphql.py`, all of the tests would pass. Now, if we did that, this one would fail. It makes a graphql call to `getSlateLineup` and ensures that the response matches client expectations.

## My Goals, Expressed as Questions for the Reviewer:

1. **Is the test clear?** - i.e. if this started failing, could another developer open it up and figure out what it's supposed to be doing?
2. **Is it instructive?** - i.e. if this started failing, what's the probability that it's signal (something in the implementation broke) vs noise (the test is flaky or the tools I used for testing had a backward-breaking change)


app/models/slate_lineup.py
```diff
@@ -22,7 +22,8 @@ class SlateLineupModel(BaseModel):
 
     @staticmethod
     @xray_recorder.capture_async('models_slate_lineup_get_slate_lineup')
-    async def get_slate_lineup_with_fallback(slate_lineup_id: str, user_id: Optional[str] = None,
+    async def get_slate_lineup_with_fallback(slate_lineup_id: str,
```
chelseatroy, line 25: This is a formatting change. All the arguments were on their own line except for one, so I gave that argument its own separate line.

tests/functional/graphql/test_get_slate_lineup.py
```diff
@@ -0,0 +1,84 @@
...
+            #             {'description': 'The best longform journalism and essays popular in Pocket'},
+            #             {'description': 'Great stories that stand the test of time'}
+            #         ]
+            #         }
+            #     }
+            # }
+
+    def populate_candidate_sets_table(self):
+        self.candidate_set_table.put_item(Item={
+            'id': '303174fc-a9ff-4a51-984a-e09ce7120d18',
```
jeshuaborges, line 66: Its interesting seeing these seeds. It looks like in a lot of ways we are testing the config. Because these lists are define elsewhere. Testing is always weird where some data is in config and the other is in the database. Out of scope of this PR, it makes me wonder if we should be looking more towards a soft config and moving slates and slate lineups configs to db and treating them as more mutable values depending on their environment.

jeshuaborges (commented):
I am not a reviewer of PRs so I am not directly approving. The test makes sense to me and its easy to understand. I did have some questions as far as conventions and general thoughts. Do with them what you will.

tests/functional/graphql/test_get_slate_lineup.py
```diff
@@ -0,0 +1,84 @@
...
+            # because our clients depend on this structure.
+            slate_list = executed.get('data').get('getSlateLineup').get('slates')
+
+            # Testing that multiple slates come back
+            assert len(slate_list) > 1
+
+            # Testing that they all have a description
+            assert all('description' in slate for slate in slate_list)
+
+            # FOR YOUR REFERENCE, the response JSON should look something like this:
```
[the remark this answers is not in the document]
chelseatroy: Exactly: I imagined the purpose of the test to be to check that a valid request produces a response of the shape that we expect. But if it fails in the future because, for example, we change the wording on one of the slate descriptions, I think that would raise its noise to signal ratio without adding diagnostic benefit.

tests/functional/graphql/test_get_slate_lineup.py
chelseatroy: I think that might make sense. I'd love for us to do what we can to treat "stuff that's likely to change" as data and "stuff that's likely to not change" as code/infra, to the degree we can.

mmiermans: Given that this is a very high-level API test, of which there won't be many, I think it will be easy to change the test if the lineup in question changes. So I like the approach taken here.

Tests in [TestSlateLineupModel](https://github.com/Pocket/recommendation-api/blob/main/tests/functional/models/test_slate_lineup_model.py#L94-L95) patch the config to create fixtures for all of the data stored in JSON configs, such that tests will pass even when slates and lineups change. The way fixtures are defined in TestSlateLineupModel (60 lines of code at the top of the file) makes the code hard to understand and change, but this could be improved by using a library for organizing fixtures such as [PyTest fixtures](https://docs.pytest.org/en/latest/explanation/fixtures.html).

mmiermans (approved):
I was surprised when I got to the end sooner than expected. Nice work on writing this test in few lines, and keeping it readable!

tests/functional/graphql/test_get_slate_lineup.py
```diff
@@ -0,0 +1,84 @@
+from graphql.execution.executors.asyncio import AsyncioExecutor
+from graphene.test import Client
+from fastapi.testclient import TestClient
+
+from app.graphql.graphql import schema
+from app.main import app, load_slate_configs
+from tests.functional.test_dynamodb_base import TestDynamoDBBase
+
+from unittest.mock import patch
+from collections import namedtuple
```
chelseatroy, line 10: Wicked. We're almost ready to merge. 

@gkatre what do you think of this test? Do you think that this reads clearly enough to explain the behavior of the `getSlateLineup` endpoint?

gkatre (approved):

PR #626, opened 2022-02-01
Add new tab slate

chelseatroy:
I have some questions about how some of these fields are used, so I'll plan to ping @lfishhead when he is out of meetings to confirm that a) this is all that needs done and b) these are the right things to put in these fields.


app/json/slate_configs.json
```diff
@@ -89,6 +89,24 @@
       }
     ]
   },
+  {
+    "id": "f99178fb-6bd0-4fa1-8109-cda181b697f6",
+    "displayName": "New Tab Recommendations",
```
chelseatroy, line 94: I didn't mention en-us in the displayName or the description because if we're displaying those somewhere, readers do not need to know about feeds. I put it in the internalDescription instead. However, that means the slate titles and descriptions for all 5 feeds will be similar/the same and the internalDescription will be the differentiator. Is that in line with our desired conventions?

app/json/slate_configs.json
```diff
@@ -89,6 +89,24 @@
       }
     ]
   },
+  {
+    "id": "f99178fb-6bd0-4fa1-8109-cda181b697f6",
```
chelseatroy, line 93: I literally went to a uuid generator and got this. These aren't indexed in a db or anything are they? I'm 99% sure know but I'd like confirmation from The Data Guy

app/json/slate_configs.json
```diff
@@ -89,6 +89,24 @@
...
+    "displayName": "New Tab Recommendations",
+    "description": "A selection of content for display on the Firefox new tab",
+    "internalDescription": "en-us New Tab recommendation items for today",
+    "experiments": [
+      {
+        "description": "default",
+        "candidateSets": [
+          "493a5556-9800-449f-8f8c-c27bb6c8c810"
```
chelseatroy, line 101: Same candidate set as Editor's Picks on purpose

app/json/slate_configs.json
lfishhead: the answer is yes

app/json/slate_configs.json
lfishhead: Joel may be the main consumer of the displayName

app/json/slate_configs.json
```diff
@@ -89,6 +89,24 @@
...
+        ],
+        "rankers": [
+          "top30-items",
+          "thompson-sampling-28day"
```
lfishhead: you can remove thompson-sampling if you want predictable item ordering within the slate

lfishhead (approved):

app/json/slate_configs.json
```diff
@@ -89,6 +89,74 @@
       }
     ]
   },
+  {
+    "id": "f99178fb-6bd0-4fa1-8109-cda181b697f6",
+    "displayName": "New Tab Recommendations",
```
mmiermans, line 94: Matt is correct. Firefox supplies these strings internally.

The string that's used by Firefox today is this:
```suggestion
    "displayName": "Recommended by Pocket",
```

mmiermans (approved):
:+1: :ship:

PR #627, opened 2022-02-01
Create type CorpusRecommendation

jeshuaborges:
# Goal

Demonstrate how we are going to transition to a new recommendation type backed by the curated corpus.

## Todos
- [x] Demo graphql schema

## Review
- [ ] Schema @bassrock @jpetto @chelseatroy 

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/BACK-1256?focusedCommentId=66537&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-66537

## Implementation Decisions

Factors at play:
1. We are going to transition [all recommendations to come from the corpus](https://pocket.slack.com/archives/CTLKW5WCF/p1643823032942869). 
2. By creating new properties we will be ensuring that CorpusItem is always available for recommendations from this code path. This will allow clients to request much fewer parameters in order to render their item.
3. We are deprecating item_id. So all the legacy Slate properties can be deleted as we migrate all systems to the new corpus properties.

Example client query with new schema:

```graphql
query getCorpusSlate {
  getCorpusSlate(id: "[UUID]") {
    name
    recommendationResult: {
       recommendations(10) {
         id
         corpusItem: {
           title
           url
           exerpt
           imageUrl
           publisherName
         }
       }
    }
  }
}
```

[Example client query with current schema from android](https://studio.apollographql.com/graph/pocket-client-api/operations?metric=requestRate&query=dbf5d3bad52486c7e9a3227085f4eecf22b004c6&queryName=getSlateLineup&range=lastDay&variant=current):
```graphql
fragment AuthorFields on Author {
  id
  name
  url
}

fragment CuratedInfoFields on CuratedInfo {
  excerpt
  imageSrc
  title
}

fragment DomainMetadataFields on DomainMetadata {
  logo
  logoGreyscale
  name
}

fragment ImageFields on Image {
  caption
  credit
  height
  imageId
  src
  width
}

fragment ItemFields on Item {
  ampUrl
  authors {
    ...AuthorFields
  }
  domain
  domainMetadata {
    ...DomainMetadataFields
  }
  encoding
  excerpt
  givenUrl
  hasImage
  hasVideo
  images {
    ...ImageFields
  }
  isArticle
  isIndex
  itemId
  mimeType
  normalUrl
  resolvedId
  resolvedUrl
  syndicatedArticle {
    ...SyndicatedArticleFields
  }
  title
  topImageUrl
  videos {
    ...VideoFields
  }
  wordCount
}

fragment PublisherFields on Publisher {
  name
  url
}

fragment RecommendationFields on Recommendation {
  curatedInfo {
    ...CuratedInfoFields
  }
  id
  item {
    ...ItemFields
  }
}

fragment SlateFields on Slate {
  description
  displayName
  experimentId
  id
  recommendations {
    ...RecommendationFields
  }
  requestId
}

fragment SyndicatedArticleFields on SyndicatedArticle {
  publisher {
    ...PublisherFields
  }
  slug
}

fragment VideoFields on Video {
  height
  length
  src
  type
  vid
  videoId
  width
}

query getSlateLineup() {
  getSlate(
    count: 10
    slateId: "[UUID]"
  ) {
    ...SlateFields
  }
}
```


schema.graphql
```diff
@@ -177,6 +177,30 @@ type Recommendation @key(fields: "item { itemId }") {
...
+    """
+    url: String!
+
+    """
+    Corpus item id.
+    """
+    corpusItemId: String!
+}
+
+union Recommendation = ParserRecommendation | CorpusRecommendation
```
jeshuaborges, line 202: Apparently you cannot change a type from a native type to a union type without possibly impacting clients.

schema.graphql
```diff
@@ -19,6 +19,9 @@ type Query {
         recommendationCount: Int = 10
     ): Slate
 
+    # TODO: Is there a better name for `RecommendationResult`?
+    getCorpusSlate(slateId: String!): CorpusSlate
```
jeshuaborges, line 23: We build new types so that we can deliver `CorpusItemRecommendation` which keys its content on `CorpusItem` from the curation tools.

schema.graphql
```diff
@@ -40,6 +43,52 @@ type Query {
...
+    A unique identifier for this specific recommendation across all recommendations. This identifier is generated by the
+    a machine learning system making recommendation decisions. This field can be joined with recommendation decisions to
+    aggregate more meta data about the decision.
+    """
+    id: ID!
+
+    """
+    Corpus item id.
+    """
+    corpusItemId: ID!
```
jeshuaborges, line 87: Corpus management tool will have to extend CorpusItemRecommendation to include `corpusItem: CorpusItem!`.

bassrock: > Would it be better, is it possible, to define the corpusItem property here?

This should only define what it knows about which is the corpusItemId. Then we let the subgraphs that know what to do with that extend and implement it. 

The key thing to remember is clients see the built graph and this will not matter to them. 


Now.... So now that I say this, i then scroll down and see Recommendation implements Item... and somehow resolves.... so we could also do it that way and CuratedCorpus never needs to know about recomendation api. I guess it depends on how we want to implement our graphs.

schema.graphql
```diff
@@ -40,6 +43,52 @@ type Query {
     ): SlateLineup
 }
 
+type CorpusSlate @key(fields: "id") {
+    id: ID!
+    name: String!
+    description: String!
+    experiments: [CorpusSlateExperiment!]!
+    recommendationResult: CorpusSlateRecommendationResult
```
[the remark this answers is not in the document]
jeshuaborges: A slate is different than its results. A slate has many static properties while the results may be specific to the user and the current request.

jeshuaborges:
Closing this PR in favor of a higher level conversation in https://github.com/Pocket/spec/pull/133#pullrequestreview-886044727

PR #633, opened 2022-02-04
Add ranker for Firefox New Tab [FFRECSV2-125]

mmiermans:
# Goal
- Get engagement metrics from a Feature Group by recommendation id.
- Allow the Thompson sampling ranker to use these metrics.

## QA
- [x] QA in Pocket-Dev, by copying a candidate set to DynamoDB, inserting dummy data in Feature Group, and checking that higher CTR items end up on top on average.
- [x] Functional test for getting metrics from a Feature Group
- [x] Functional test for using Thompson sampling ranker with the new metrics schema
- [x] Functional test for getting a slate configured to rank using `firefox-thompson-sampling-15minute`

## Deployment
- [x] Feature Group is created in Pocket-Dev
- [x] Feature Group is created in production

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/FFRECSV2-125

## Implementation Decisions
- Depending on performance requirements, getting 30+ records from a Feature Group in individual requests (even if they happen concurrently) might not be performant enough. I documented the necessary steps to batch these.
- I'm not sure if prefixing with `firefox_new_tab_` resulted in the best names. Happy to rename if anyone has a suggestion.
- I changed the Thompson sampling test from using loops to a parameterized test. It was getting a little out of hand.
- We've recently started to create shared policies such as `PocketDataProductReadOnly`. I added fine-grained permissions instead to support deployment in both Pocket-Dev and production. Additionally, RecommendationAPI is fairly stable, and probably won't benefit as much from having the shared policy as services like Metaflow or Prefect.

## PR walk-through
Try downloading these, because video resolution streaming from Google Drive can be pretty bad:
- [Part 1](https://drive.google.com/file/d/1E2QdHwPjIM3PTmIbH_mByWwMjzyAv7Q3/view?usp=sharing): Everything up to the Thompson sampling tests.
- [Part 2](https://drive.google.com/file/d/1LB3ERuqGzJcOBrHFlGxDGVG1z7P7Nn7d/view?usp=sharing): Thompson sampling test


.aws/src/main.ts
```diff
@@ -295,6 +295,15 @@ class RecommendationAPI extends TerraformStack {
                         ],
                         effect: 'Allow'
                     },
+                    {
```
jeshuaborges, line 298: Would it be better to attach the managed policy `arn:aws:iam::996905175585:policy/PocketDataProductReadOnly`?

mmiermans: I considered that, but `PocketDataProductReadOnly` is currently only available in our production account, and I like that RecommendationAPI is able to deploy to both Pocket-Dev and production.

app/models/metrics/firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,102 @@
...
+
+# Unfortunately boto3 doesn't use types for feature store records at the moment, so we define one ourselves.
+# For example records see tests/assets/json/firefox_new_tab_engagement.json
+FeatureStoreRecordType = List[Dict[str, str]]
+
+
+class FirefoxNewTabMetricsFactory():
+    _dynamodb_endpoint: str = None
+    _dynamodb_table: str = None
+    _PRIMARY_KEY_NAME: str = 'ID'
```
mmiermans, line 22: Can this be removed?

app/rankers/algorithms.py
```diff
@@ -110,39 +118,38 @@ def thompson_sampling(
...
-            # Slates are keyed on their slate id, in this case the id field of the slate config model
-            # Similarly these are parsed as the prefix of the primary key in the slate metrics table
-            clickdata_id = rec.id
-
-        d = metrics.get(clickdata_id)
-        if d:
-            # TODO: Decide how many days we want to look back.
-            clicks = max(getattr(d, opens_column) + alpha_prior, 1e-18)
+            metrics_model = metrics[rec.id]
+        except (AttributeError, KeyError):
```
mmiermans, line 139: Change this to only KeyError

app/rankers/algorithms.py
```diff
@@ -110,39 +118,38 @@ def thompson_sampling(
...
+            # Legacy recommendations are keyed on item_id.  Note that the metrics model grabs the item_id
+            # when it parses the clickdata by splitting the primary key in dynamo.
+            metrics_model = metrics.get(rec.item.item_id, None)
```
mmiermans, line 142: Catch the AttributeError here

tests/functional/models/test_firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,130 @@
...
+
+import app.config
+from app.models.metrics.firefox_new_tab_metrics_factory import FirefoxNewTabMetricsFactory
+
+
+class FeatureStoreClient:
+    def __init__(self, **kwargs):
+        pass
+    async def get_record(self, **kwargs):
+        pass
```
mmiermans, line 20: Unused code?

tests/functional/models/test_slate_model.py
```diff
@@ -1,31 +1,64 @@
...
+firefox_slate_config_id = 'test-firefox-slate_lineup-config-id'
+firefox_slate_experiment = SlateExperimentModel('test-firefox-ex', 'test-ex-firefox-desc',
+                                                ['firefox-thompson-sampling-1day'], ['test-candidate-id'])
+firefox_slate_config_model = SlateConfigModel(firefox_slate_config_id, 'test-firefox-slate', 'test-firefox-desc',
+                                              experiments=[firefox_slate_experiment])
+
 
 class TestSlateModel(TestDynamoDBBase):
     async def asyncSetUp(self):
         await super().asyncSetUp()
```
mmiermans, line 47: Remove

tests/functional/models/test_slate_model.py
```diff
@@ -56,39 +93,35 @@ async def test_get_slates_from_slate_configs_without_recs(self):
...
+            get_firefox_metrics_mock
+    ):
+        self._put_candidates(4)
+
+        slates = await SlateModel.get_slates_from_slate_configs([firefox_slate_config_model], None)
+
+        assert slates[0].id == firefox_slate_config_id
+        assert len(slates[0].recommendations) == 4
+        get_firefox_metrics_mock.assert_called_once()
+        firefox_thompson_sampling_mock.assert_called_once()
```
mmiermans, line 120: Assert that `firefox_thompson_sampling_mock` is called with the right metrics

jeshuaborges (commented):
I ran out of time to complete a full review. Here are some notes.

.aws/src/main.ts
chelseatroy: I like it, too—until and unless we make a feature decision in the future that requires Recs API to write data somewhere (for example if we end up needing to record what we sent in which order to which user such that we can more effectively investigate cases of misinformation recommendations in the future). 

NOT A NEAR TERM CONCERN. But wanted to mention it apropos of our prior discussions about metaflow, feature groups, and prefect that resulted in our decisions to stay prod-only with our data writes.

app/json/slate_configs.json
```diff
@@ -101,7 +101,8 @@
           "493a5556-9800-449f-8f8c-c27bb6c8c810"
         ],
         "rankers": [
-          "top30-items"
+          "top30-items",
+          "firefox-thompson-sampling-1day"
```
chelseatroy, line 105: I actually think it's fine to leave this in here, and as a matter of fact, to add it to the three slates mentioned in the config _below_ this one.

Why: that's all 4 new tab slates and I'd like us to default to those all having consistent behavior unless we've got a good reason to make them different.

app/models/metrics/firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,95 @@
...
+from app.models.metrics.firefox_new_tab_metrics_model import FirefoxNewTabMetricsModel
+from app.models.metrics.metrics_model import MetricsModel
+
+
+# Unfortunately boto3 doesn't use types for feature store records at the moment, so we define one ourselves.
+# For example records see tests/assets/json/firefox_new_tab_engagement.json
+FeatureStoreRecordType = List[Dict[str, str]]
+
+
+class FirefoxNewTabMetricsFactory():
```
chelseatroy, line 19: Apropos of your request for input in the video, I think this name and the other name are both fine for now. Eventually we may want to name these for their sources (Dynamo and FeatureGroups) if we end up using them—and especially this one—to populate additional surfaces as the app grows, but that's not a thing we need to do right now.

app/models/metrics/firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,95 @@
...
+        return f'{config.ENV}-firefox-new-tab-engagement-v{cls._FEATURE_GROUP_VERSION}'
+
+    def _parse_from_record(self, record: FeatureStoreRecordType) -> FirefoxNewTabMetricsModel:
+        """
+        Converts a FeatureGroup record to a Pydantic FirefoxNewTabMetricsModel object
+        :param record: dictionary containing all required keys for FirefoxNewTabMetricsModel
+        :return: Parsed FirefoxNewTabMetricsModel
+        """
+        # Pydantic model uses lowercase keys.
+        features = {feature['FeatureName'].lower(): feature['ValueAsString'] for feature in record}
```
chelseatroy, line 61: I would do this in `parse_obj` probably, but I don't feel strongly

app/models/metrics/firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,95 @@
...
+
+        :param metrics_keys: Feature record IDs to query
+        :return: List with all metrics that were successfully queried from the Feature Store.
+                 If metrics are missing from the output that means the feature group did not find a match, probably
+                 because the recommendation is too recent.
+        """
+        metrics = []
+
+        async with aioboto3.client('sagemaker-featurestore-runtime') as featurestore:
+            # TODO: Update aioboto3 to v9 to improve performance with featurestore's batch_get_record.
```
chelseatroy, line 77: I know you said you feel that using the batch query would be a premature optimization, but given that the solution might just be updating this library, I don't think it would be out of line to take, say, a one hour time box and give it a try. Were I doing this, the important thing would be for me to make a call at 45 or 50 minutes. Will I for sure be done in 10-15 mins? If not, I'd go ahead and write up a ticket with findings for to come back to this later.

However I defer to @jeshuaborges on this

app/models/metrics/firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,95 @@
...
+            #       - boto3 1.17.92 introduces BatchGetRecord
+            #       - aioboto3 8.3.0 pins aiobotocore[boto3]==1.2.2
+            #       - aiobotocore 1.2.2 pins boto3 1.16.52
+            promises = [featurestore.get_record(
+                FeatureGroupName=self.get_feature_group_name(),
+                RecordIdentifierValueAsString=metrics_key,
+                FeatureNames=self._FEATURE_NAMES
+            ) for metrics_key in metrics_keys]
+
+            responses = await gather(*promises)
```
chelseatroy, line 87: To answer your question from the video: the number @lfishhead has historically used in these situations is 3 times the requested number to ensure enough after duplicates and blocked items. I don't think we need that many here, particularly given that this ranker is gonna get used for curated new tab recommendations that should already exclude those things, but would be interested in his input on the multiplier here.

app/models/recommendation.py
```diff
@@ -12,10 +12,11 @@
 # Needs to exist for pydantic to resolve the model field "item: ItemModel" in the RecommendationModel
 from app.graphql.item import Item
 from app.models.candidate_set import candidate_set_factory
+from app.models.metrics.firefox_new_tab_metrics_factory import FirefoxNewTabMetricsFactory
 from app.models.metrics.recommendation_metrics_factory import RecommendationMetricsFactory
 from app.models.item import ItemModel
 from app.models.slate_experiment import SlateExperimentModel
-from app.rankers import get_ranker, THOMPSON_SAMPLING_RANKERS
+from app.rankers import get_ranker, THOMPSON_SAMPLING_RANKERS, FIREFOX_THOMPSON_SAMPLING_RANKERS
```
chelseatroy, line 19: Naming stuff like `A_THING` and `MORE_QUALIFIERS_A_THING` to me suggests that the second one is a subset of the first one. If they are peers (as we have here with `THOMPSON_SAMPLING_RANKERS` and `FIREFOX_THOMPSON_SAMPLING_RANKERS`), we can indicate that by putting qualifiers on both of their names.

app/rankers/algorithms.py
```diff
@@ -16,6 +17,11 @@
 DEFAULT_ALPHA_PRIOR = 0.02
 DEFAULT_BETA_PRIOR = 1.0
 
+# These values were taken from: https://github.com/Pocket/feed-machine/blob/master/resources/models/b-0085-15k-age-05-syn-score-explore-locales-rr.json#L4-L5
+# They were experimentally derived in 2018: https://getpocket.atlassian.net/browse/P18-616
+DEFAULT_FIREFOX_BETA_PRIOR = 15600
```
chelseatroy, line 22: Agreed with your comment in the video: this looks low, but right now I'm not that worried about starting at the perfect value.

tests/functional/models/test_firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,121 @@
...
+            # Raise an exception if the feature group is not found.
+            # In reality FeatureStore raises a ClientError, but the BotoCoreError is easier to construct for testing.
+            raise BotoCoreError()
+
+        full_record = self._firefox_new_tab_engagement_record_by_id.get(record_identifier_value_as_string, {})
+        # Only return requested features to check that a valid model can be constructed from the requested features.
+        filtered_record = [feature for feature in full_record if feature['FeatureName'] in feature_names]
+        return {'Record': filtered_record}
+
+    async def test_get_existing_records(self, monkeypatch):
```
chelseatroy, line 62: When I try to run the tests, they're getting skipped in my editor and I get a warning about not having a suitable plugin installed (I looked at my pip environment and pytest-asyncio was there). I didn't spend time trying to get them running tonight, but wanted to flag. Is there another step needed here to get these running, a run configuration or something?

tests/functional/models/test_slate_model.py
```diff
@@ -56,39 +91,43 @@ async def test_get_slates_from_slate_configs_without_recs(self):
...
 
     async def test_get_slates_from_slate_configs_with_recs(self):
+        self._put_candidates(1)
+
         slates = await SlateModel.get_slates_from_slate_configs([slate_config_model], None)
 
         assert slates[0].id == slate_config_id
         assert len(slates[0].recommendations) == 1
 
+    @patch('app.models.metrics.firefox_new_tab_metrics_factory.FirefoxNewTabMetricsFactory.get',
```
chelseatroy, line 101: It's not great that we have to patch our own stuff in tests. Outside methods, sure, but maybe there's a path forward with dependency injection instead.

I'd be prepared to call that out of scope for this specific ticket, though.

tests/unit/rankers/test_algorithms.py
```diff
@@ -149,54 +150,61 @@ def test_block_item_using_blocklist_param(self):
...
+            )
+        # Invalid trailing_period for Firefox New Tab metrics
+        with pytest.raises(ValueError):
+            thompson_sampling(
+                generate_recommendations(['000-999']),
+                metrics=generate_firefox_metrics(['000-999']),
+                trailing_period=7  # MetricsModel has 7 day trailing period, but FirefoxNewTabMetricsModel does not.
+            )
+
+    @pytest.mark.parametrize("thompson_sampling_function,metrics", [
```
chelseatroy, line 200: I'm not usually a fan of parametrized tests, but this is _exactly_ the use case where I think they make sense 🙂

chelseatroy (commented):
See comments. I made some minor changes that I thought would be uncontroversial in an add-on commit to this PR!

tests/functional/models/test_firefox_new_tab_metrics_factory.py
mmiermans: Did you run the tests using docker-compose? [The development setup docs](https://github.com/Pocket/recommendation-api/blob/main/docs/development.md#runningdebugging-app-tests-in-pycharm) describe how to run the tests.

tests/functional/models/test_slate_model.py
mmiermans: Agreed.

app/rankers/algorithms.py
```diff
@@ -110,39 +118,42 @@ def thompson_sampling(
...
-            # Similarly these are parsed as the prefix of the primary key in the slate metrics table
-            clickdata_id = rec.id
-
-        d = metrics.get(clickdata_id)
-        if d:
-            # TODO: Decide how many days we want to look back.
-            clicks = max(getattr(d, opens_column) + alpha_prior, 1e-18)
+        # When metrics data no longer keyed on item_id, we can simple do `metrics_model = metrics.get(rec.id)`.
+        if rec.id in metrics:
+            metrics_model = metrics[rec.id]
```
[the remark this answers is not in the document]
mmiermans: I also don't like that we're passing the join column as unnamed dictionary keys. I'd prefer it being a field on the model.

app/models/metrics/firefox_new_tab_metrics_factory.py
mmiermans: Ticket for this change: https://getpocket.atlassian.net/browse/FFRECSV2-145

app/models/metrics/firefox_new_tab_metrics_factory.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        Get engagement metrics for a Firefox New Tab slate experiment.
+        TODO: This function should simply accept 'recommendation_ids', and we should generate these ids elsewhere.
+        :param recommendation_ids: GUIDs uniquely represent a scheduled run of a corpus item in a slate experiment.
+        :return: dictionary of FirefoxNewTabMetricsModel objects keyed on recommendation id
+        """
+        # Keys are namespaced by the slate that we are getting data from. First put them in a set to ensure unique keys.
+        metrics = await self._query_metrics(recommendation_ids)
+        if not metrics:
+            logging.error(f"No Firefox New Tab metrics for recommendation_ids={recommendation_ids}")
```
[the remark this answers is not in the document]
jeshuaborges: I dont think you's statement is correct. At some point, new `recommendation_ids` should appear. If they are all new, say when a system is just started then metrics will not be available for those `recommendation_ids`.

tests/functional/models/test_firefox_new_tab_metrics_factory.py
chelseatroy: I did use that setup. It turns out I gotta do `build` and `up` in the Pycharm terminal—the regular terminal doesn't work for this, which is where I normally do command line stuff.

chelseatroy (approved):
This looks ready to me! Thanks for making those changes.

jeshuaborges (approved):
Approving without complete review. I dont want to block progress. Thanks for pushing this through mathijs.

PR #636, opened 2022-02-04
Put URL in the Candidate Set

chelseatroy:
Add validation for presence of url, the contemporary pocket primary key of record, in candidates before sticking them in DynamoDB.

As of this PR metaflow will add url to the SQS Queue: https://github.com/Pocket/dl-metaflow-jobs/pull/84

This PR is to the lambda that gets the candidate sets off the SQS Queue and puts them into DynamoDB. Now they will have their urls in there.

**Benefits**
- See aforementioned PR.

**Risks**
- We need to check if there is some kind of row size limit in Dynamo DB, and if so how close we are to it, and if we're close, whether adding a url to each of our candidates puts us over that limit. My hope on this is 'no' but if the answer is 'yes' things get more complicated.

### DIFF
--- a/aws_lambda/sqs_handler.py
+++ b/aws_lambda/sqs_handler.py
@@ -83,6 +83,7 @@ def _validate_candidate_set(candidate_set: Dict[str, Any]):
 
 def _validate_candidate(candidate: Dict[str, Any]):
     _validate_dict_value_type(candidate, 'item_id', int)
+    _validate_dict_value_type(candidate, 'url', str)
     _validate_dict_value_type(candidate, 'publisher', str)
 
 


aws_lambda/sqs_handler.py
```diff
@@ -83,6 +83,7 @@ def _validate_candidate_set(candidate_set: Dict[str, Any]):
 
 def _validate_candidate(candidate: Dict[str, Any]):
     _validate_dict_value_type(candidate, 'item_id', int)
```
jeshuaborges, line 85: I would expect that `item_id` is not required when `url` is provided.

aws_lambda/sqs_handler.py
```diff
@@ -83,6 +83,7 @@ def _validate_candidate_set(candidate_set: Dict[str, Any]):
...
+    _validate_dict_value_type(candidate, 'url', str)
```
mmiermans, line 86: This might address Jesh's comment:
```suggestion
    try:
        _validate_dict_value_type(candidate, 'url', str)
    except AssertionError:
        _validate_dict_value_type(candidate, 'item_id', int)
```
^ Untested

mmiermans (commented):
- The tests for this Lambda are failing on the `expires_at` validation. I'll see why that is.
- Once that's fixed, it should be pretty easy to add a test to this PR by copying one of the existing assertions for `item_id`: https://github.com/Pocket/recommendation-api/blob/main/aws_lambda/tests/test_sqs_lambda.py#L121-L125

mmiermans:
Here's a change to fix the failing `test_lambda` check: https://github.com/Pocket/recommendation-api/pull/642

jeshuaborges:
This work has been dropped. If we want to circle back we can revisit.

PR #641, opened 2022-02-07
Key on, collect item with, and surface url from the recs api

chelseatroy:
This would cover the API "frontend" changes for the changeover from itemID to url for which we changed the lambda here: https://github.com/Pocket/recommendation-api/pull/636

- I'd need to work with Mathijs on the appropriate way to test that this would work, particularly given that it possesses the federated component where it fetches information from the graph.
- This would be the final of the 3 PRs (the one referenced above and the one it references in Metaflow) to surface URL alongside itemID and use it where possible.

The other option besides these 3 PRs is to create a whole separate flow with a separate endpoint, as per @jeshuaborges' suggestion. I think that the work to implement _this_ attempt is useful context for trying _that_, which I can start on tomorrow, Tuesday, February 8.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,6 +1,6 @@
 type Query {
     """
-    Get the recomendations for a specific topic
+    Get the recommendations for a specific topic
     """
     getTopicRecommendations(slug: String!, algorithmicCount: Int = 20, curatedCount: Int = 5): TopicRecommendations  @deprecated(reason: "Use `getSlateLineup` with a specific SlateLineup instead.")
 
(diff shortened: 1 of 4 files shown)


schema.graphql
```diff
@@ -234,9 +238,7 @@ type SlateLineup {
...
 
-extend type Item @key(fields: "itemId") {
-    """
-    The ID of the item that is used by apollo federation to reference the Item entity.
-    """
-    itemId: String! @external
-}
+extend type Item @key(fields: "givenUrl") {
+  "key field to identify the Item entity in the Parser service"
+  givenUrl: Url! @external
```
jeshuaborges, line 243: I am under the impression that most of the time the URL you are providing from Metaflow processes is the `resolved_url`.

lfishhead: the `resolved_url` that we have in metaflow is generally from elasticsearch (and sometimes the metadata `item_store`).  So I believe this is the `resolved_url` which @gkatre  could likely verify.

jeshuaborges:
I believe that we all agreed that we are going to take a different strategy for this feature.

jeshuaborges:
This PR is no longer relevant.

PR #648, opened 2022-02-09
Create config to debug in Dev environment

mmiermans:
# Goal
Run the app locally while connecting to AWS resources in Pocket-Dev. This proved helpful while writing code for Sagemaker Feature Groups, which is currently not supported by LocalStack.

## Implementation Decisions
- I considered it out-of-scope to document how to get data into Pocket-Dev, because there's currently not a clear answer for that.
- I haven't yet created a configuration to (safely) run RecommendationAPI against production resources, so I added a warning with guidance on how to do that.
- Made the image urls relative to this markdown doc so they'll render in PyCharm.


kkyeboah (approved):
I see https://github.com/Pocket/recommendation-api/pull/642 fixes the lambda test failing here. I haven't tested this but the docs look thorough

.docker/local-dev.env
```diff
@@ -0,0 +1,13 @@
+# Disable XRay locally
+AWS_XRAY_SDK_ENABLED=false
+
+AWS_DYNAMODB_ENDPOINT_URL=https://dynamodb.us-east-1.amazonaws.com
```
gkatre, line 4: Do you need to set `AWS_PROFILE=pocket-dev-PocketSSODataLearning` in this file?

gkatre (approved):
@mmiermans just one comment. Otherwise look good! 👍

.docker/local-dev.env
mmiermans: I set the environment variable in the run configuration (line 8):
```xml
<env name="AWS_PROFILE" value="pocket-dev-PocketSSODataLearning" />
```
Not really sure why I chose to do that. Either way would be fine with me.

PR #650, opened 2022-02-09
Expose a GQL Endpoint on Recs API that Fetches Items from Corpus Tools API

chelseatroy:
# Goal

- When a client requests rankedCorpusItems from the client api with an id pertaining to the surface that they are for,
- the recs api fetches a sequence of items from the Curation tools API, requesting only their ids.
- It then returns a response of the requested shape to the client.
- In addition, the client can fetch other data about these items, which the graph can add based on the id.

Screen Shot 2022-02-14 at 6 57 09 PM

## What's still in progress?

- The curation tools API endpoint is published, but it doesn't return "real" data yet. So instead of returning our response from curation tools, the API client in this PR requests that data, ensures that the request succeeded, and then sends back some fabricated data to allow us to make sure that this request works all the way through.
- Federation does work (I can request additional details about items through the dev client api). There are no items coming through yet, though, for the graph to hydrate. 

Tickets:
* Link to JIRA tickets: https://getpocket.atlassian.net/browse/FFRECSV2-131

## Implementation Decisions

- The false data: this allowed us to perform the integration and make sure the request worked without relying on the curation tools API to be fully set up yet for us to move forward. I will make a ticket to revisit this integration once curation tools API is ready to go.
- The next step after this ticket will be to take the items from the curation tools response, fetch engagement data for them, and then put them in a new order with thompson sampling. Much of that work happens in this ranker PR:https://github.com/Pocket/recommendation-api/pull/633 . I'll be integrating these two things together next.


app/api_clients/curation_api_client.py
```diff
@@ -0,0 +1,38 @@
...
+                items(date: "{start_date}") {{
+                  id
+                }}
+            }}        
+        }}
+        """
+        # Sounds like we want this to go directly to corpus API.
+        # Let's talk to backend about this
+        url = 'https://client-api.getpocket.com'
+        response = requests.post(url, json={'query': query})
```
bassrock, line 30: Can we add a few identifying headers for Apollo Studio? This will let future us know what client is making the request.


Here's a usage in metaflow example https://github.com/Pocket/dl-metaflow-jobs/blob/210a14af2259ac4b26438d20b76fd93b3e53f18f/jobs/recs_api.py#L13-L16


We have some docs on this at https://github.com/Pocket/spec/tree/main/query/client-api#query-naming, but i know that isn't very discoverable, please bear with us as we look to find a better place for this.

app/api_clients/curation_api_client.py
```diff
@@ -0,0 +1,38 @@
...
+
+class CurationAPIClient(object):
+    @classmethod
+    async def get_ranked_corpus_items(cls, corpus_id: str, start_date: str=None, user_id=None):
+        ranked_corpus_items_id = "NEW_TAB_EN_US"
+        if not start_date:
+            start_date = date.today().strftime("%Y-%m-%d")
+
+        query = f"""
+        query Query {{
```
bassrock, line 18: Can we name the query, this is the name that will show up in Apollo Studio when we look at metrics segmented by clients.

(metaflow example https://github.com/Pocket/dl-metaflow-jobs/pull/80)

bassrock (commented):
👋  hi passerby here. I saw this pr and got excited. I left 2 comments on the call you make to client api.  They don't need to block or anything but it would be valuable to get them in to view metrics in Apollo Studio.

app/api_clients/curation_api_client.py
chelseatroy: Done :)

app/api_clients/curation_api_client.py
chelseatroy: Done!

mmiermans (commented):
Nice work!

Only one small thing I'd change is to remove/sanitize/validate parameters before inserting them into a GraphQL query.

I'm fine with the name 'RankedCorpusItems', but I have a preference for 'CorpusSlate' for two reasons:
    1. It comes with a description and some other metadata, so it's more than a ranked list of items. The Corpus API introduced the concept of a 'surface' that contains a list of items.
    2. I think this concept (a module with ranked/personalized recommendations) will be increasingly core to Pocket that deserves its own name. We've introduced 'slate' as a ranked list of recommendations, which RankedCorpusItems still is.
    
    
(I should add that the federated GraphQL is still a bit of magic to me, so I'm glad that Daniel took a look at that. I looks fine to me.)

app/api_clients/curation_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+        request_headers = {
+            "apollographql-client-name": "recommendations-api",
+            "apollographql-client-version": "1"
+        }
+
+        query = f"""
+        query RecsApiItemRequest {{
+            scheduledSurface(id: "{ranked_corpus_items_id}") {{
+                id
+                items(date: "{start_date}") {{
```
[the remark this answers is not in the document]
chelseatroy: Well, two things: 

1. We're calling another API here. This isn't a db query, it's an agreed upon contract with a service that is intended for outside use that itself exposes a start_date param. We are a pass-through, responsible, in theory, _only_ for putting the results of the curation tools API call in a different order with our rankers. 

2. In order to put the items in a different order, _our_ expectations from the response for this query are pretty exacting, and _our_ step will fail already if someone somehow managed to send a variable that manipulates the API contract in such a way that the response changed.

I think the risk of injection here is lowand the likelihood of a product need for clients requesting a day is high, given that we've got new tab slates in wildly different time zones querying a service that stores scheduled days in one time zone. I would rather write an api that allows clients to query for the new tab slate for a specific date (defaulting to today) than hardcode it to today over a security concern that I don't see as a risk.

chelseatroy:
Re: `CorpusSlate`, totally open to a naming discussion, but I don't want it to block this merge.

While I think names are important, I _also_ think it is easier to name things accurately after we have viewed the object in context. I could see that name changing again as I integrate ranking in this next ticket.

mmiermans (approved):
:+1:

schema.graphql
```diff
@@ -38,6 +38,35 @@ type Query {
         "Maximum number of recommendations to return in {Slate.recommendations}, defaults to 10"
         recommendationCount: Int = 10
     ): SlateLineup
+
+    """
+    Request a specific `RankedCorpusItems` by ID: ultimately replaces `getSlate`
+    """
+    getRankedCorpusItems(
+        id: String!
```
jeshuaborges, line 46: What is the ID here? Is this the same as slate id?

schema.graphql
```diff
@@ -38,6 +38,35 @@ type Query {
...
+    ): RankedCorpusItems
+}
+
+type RankedCorpusItems @key(fields: "id") {
+    id: ID!
+    description: String!
+    displayName: String!
+    experimentId: String!
+    corpusItems: [CorpusItem!]
```
jeshuaborges, line 55: Important: Every corpus item will require its own `recommendation_id` or other ID to uniquely track the engagement of this item. We can make this follow-up work when we start delivering recommendation results to the data warehouse.

jeshuaborges (commented):
A couple of comments.

chelseatroy:
@jeshuaborges This PR cannot be merged because commits `[2848cdf](https://github.com/Pocket/recommendation-api/pull/650/commits/2848cdf7166bda8b336dd2610d65350a67ea52a6)` and `[44d3d08](https://github.com/Pocket/recommendation-api/pull/650/commits/44d3d0894b0f2d7555f8609840f78fe55a8ac770)` are not verified. 

I see these are from you, so unless you object, what I'm gonna do is _remove_ that merge requirement temporarily, then merge this in, then replace the requirement.

PR #663, opened 2022-02-15
Rank CurationToolsAPI results and surface the reranked list to our GQL endpoint

chelseatroy:
# Goal

The purpose of this PR is to complete the flow of items from the Curation Tools API through the Recs API, to be ranked in the recs API, and then to be returned through our graphql endpoint. 

It implements a flow _parallel_ to the existing flow that accesses candidate sets from dynamo DB. It deliberately does so in such a way that:

- the other flow still works
- we can gradually move over to the new flow with time
- the rankers stay pretty much the same 

## Reference

Tickets:
* [FFRECSV2-138](https://getpocket.atlassian.net/browse/FFRECSV2-138?atlOrigin=eyJpIjoiZjUwYzdjYTQ5ZjkzNDFkZmEwNDg0ODg0YzI1NmE5M2YiLCJwIjoiaiJ9)

## Implementation Decisions

Many 😂. See [video!](https://drive.google.com/file/d/1m9rlwfn3S0GKQe3QEkelUFE4IIPwbNTL/view?usp=sharing)

The high-level:
- no more static methods on stuff. Instances of classes and inversion of control (dependency injection) to make components interchangeable and separately testable.
- Slate config as a python object and not a json so that structure is enforced by python objects rather than a separate json schema, changes hot reload, there's no longer a need for a registry
- Instead of raising in the event of an experiment without rankers, this flow sends through an unmodified list of items
- Instead of raising in the event of an experiment with a duplicate ranker, this flow allows duplicate rankers (it's not catastrophic, likely, or insidious to include two of the same ranker by accident)


tests/unit/data_providers/test_dispatch.py
```diff
@@ -0,0 +1,111 @@
...
+    def get(self, slate_id) -> SlateSchema:
+        #offers up this slate regardless of the slate_id sent in. We test the SlateSchema logic in that class's tests.
+        return self.schema
+
+class MockCurationAPIClient(CurationAPIFetchable):
+    mock_corpus = RankedCorpusItemsInstance(
+        id="CHELSEAS_AWESOME_CORPUS",
+        description="I mean, pretty sure the id covers it ;)",
+        corpusItems=[
+            CorpusItemModel(id="https://chelseatroy.com/2022/02/10/adding-error-productions-to-the-lox-compiler/"),
```
jeshuaborges, line 34: My understanding is that `id` in this context is a UUID. I may be wrong, if so disregard. It may be worthwhile to make the tests resemble real data for documentation.

app/data_providers/curation_api_client.py
```diff
@@ -0,0 +1,46 @@
...
+        }}
+        """
+        # Sounds like we want this to go directly to corpus API.
+        # Let's talk to backend about this
+        url = 'https://client-api.getpocket.com'
+        response = requests.post(url, headers=request_headers, json={'query': query})
+        response_body = json.loads(response.text)
+        corpus_items = response_body.get("data").get("scheduledSurface").get("items")
+
+        return corpus_items
```
jeshuaborges, line 46: Its worth updating your editor to always include a new line.

See reasons here: https://stackoverflow.com/a/5813359/2175088

tests/unit/data_providers/test_dispatch.py
```diff
@@ -0,0 +1,111 @@
...
+        ],
+    )
+
+    async def get_ranked_corpus_slate(cls, corpus_id: str, start_date: str = None,
+                                         user_id=None) -> List[CorpusItem]:
+        return cls.mock_corpus.corpusItems
+
+@pytest.mark.asyncio
+async def test_get_ranked_items__no_rankers():
+    mock_curation_api_client = MockCurationAPIClient()
```
jeshuaborges, line 51: ❤️ your approach to testing and usage of DI and mocks.

jeshuaborges (commented):
There is a lot to like about this PR. I wont enumerate them all. The tow items I like best are the the following. First you created thoughtful future forward ways of creating interfaces and enforcing those interfaces. Secondly, you were able to create a very lightweight bridge between the two implementations.

My request for reviewers. Let's not block this pr unless we have concerns about this breaking or negatively impacting the current production deployment.  If we want to have conversations about code strategies, let's break those into discrete conversations, PRs, or other follow-up work.

One general request, can we add some of the video documentation to the project readme or other project-level documentation?

Screen Shot 2022-02-22 at 8 21 04 AM

app/data_providers/curation_api_client.py
```diff
@@ -0,0 +1,46 @@
...
+        query = f"""
+        query RecsApiItemRequest {{
+            scheduledSurface(id: "{ranked_corpus_items_id}") {{
+                id
+                items(date: "{start_date}") {{
+                  id
+                }}
+            }}        
+        }}
+        """
```
mmiermans, line 38: Whenever data is inserted in code, it's good practice to escape that data, to prevent the interpreter from treating a malformed string as code. Injection vulnerability is a low risk here for several reasons, but it's still the right thing to do. I wouldn't rely on variables not being under user-control, validation happening elsewhere, or that there aren't any harmful GraphQL queries that can be run in the graph. All of those might change and are hard to verify.

1. The most reliable way to escape parameters is to let a trusted library do it for us. GraphQL support is pretty new in Python, so I'm assuming you use `requests` because there isn't a great client library available yet.
2. The next best thing is to call an escape function from a trusted library designed to escape in the given context: GraphQL strings. I couldn't immediately find one.
3. That leaves doing some custom validation/escaping ourselves. I'm reasonably confident that `json.dumps` would do the job. I'm also fine with regex validation or a Python date format function, along with a comment to the effect that this safeguards against GraphQL query injection.
```suggestion
            # Use json.dumps() to escape GraphQL strings, because we couldn't find one designed for GraphQL.
            # Better would be to find a trusted library that handles inserting parameters.
            scheduledSurface(id: {json.dumps(str(ranked_corpus_items_id))}) {{
                id
                items(date: {json.dumps(str(start_date))}) {{
                  id
                }}
            }}        
        }}
```
^ Untested code.

app/data_providers/dispatch.py
```diff
@@ -0,0 +1,73 @@
...
+        # If no rankers, return the same list in the same order that the corpus api handed us
+        ranked_items = flattened_unranked_items
+
+        # I do not check for duplicate rankers here. I know the old flow does.
+        # This error is not catastrophic (at most, it reorders items a few times)
+        # It's also not likely (an eng AND their reviewer would have to miss two identical lines next to each other)
+        # And it's not insidious (would be easy to find and verify if we suspected it was happening)
+        # So it's not worth complicating the code to check for IMO
+        for ranker in experiment.rankers:
+            ranker_kwargs = await self.metrics_client.get_engagement_metrics(ranked_items, ranker)
```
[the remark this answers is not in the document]
mmiermans: I was going to put a comment that's similar Jesh's first one, if I understand it correctly. It seems more logical to me the rankers would define their data requirements, instead of metrics clients to know what data to load for a particular ranker. This might look like a ranker using dependency injection to get an object that provides engagement metrics.

Hadn't thought about the second point yet, but that's a good point too. Should be fairly easy to gather everything first, and then map it onto the right ranker.

I don't think we have to do either of these things in this PR. Because there's just one ranker and one type of metrics, it's easy to see how they map, and there wouldn't be a performance benefit to using `gather` yet.

tests/unit/data_providers/test_dispatch.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Establish order sameness
+    assert [item.id for item in ranked_items] == [item.id for item in mock_curation_api_client.mock_corpus.corpusItems]
+
+@pytest.mark.asyncio
+async def test_get_ranked_items__one_item_set__one_ranker():
+    mock_slate_provider = MockSlateProvider()
+
+    #Add a ranker to the mock object
+    mock_slate_provider.schema.experiments[0].rankers.clear() # We need this line because there's some kind of test data interference. Not sure why.
```
mmiermans, line 70: `mock_slate_provider.schema` is a static / class variable. Could changing it interfere with other (future) tests?

mmiermans (commented):
:heart: Love the refactor! Love that we're moving to dependency injection, and away from a great number of static classes that are dependent on each other. Thanks for simplifying things, I think it was due. :bow: The video was super clear.

Only thing I'd like to see changed is to escape/validate parameters at the point where they're inserted into GraphQL.

app/data_providers/dispatch.py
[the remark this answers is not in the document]
chelseatroy: So, up until now, the flow we have been using—the one @mmiermans updated a week ago to include the firefox metrics—does this conditionally the same way the metrics client currently does.

I would love to change it. That said, one of the primary goals I had in this PR, based on anxieties expressed by this team, was to _not change_ how rankers worked. That's why I chose _this_ design decision rather than the one suggested.

I confess it's frustrating to have one choice modeled and requested and then another choice suggested after I finish up. I am tabling this idea for a future change set, and we will establish team buy-in before making it.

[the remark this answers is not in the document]
chelseatroy: The second point I agree with, but I agree with @mmiermans that it is an implementation detail for the theoretical case we do not have yet in which a ranker requires multiple sets of metrics.

tests/unit/data_providers/test_dispatch.py
chelseatroy: My understanding is that it is not. For example, the ids currently implemented for CurationToolsAPI are NEW_TAB_EN_US, NEW_TAB_EN_INTL, NEW_TAB_EN_GB, and NEW_TAB_DE_DE.

tests/unit/data_providers/test_dispatch.py
chelseatroy: Goood catch! Will adjust.

app/data_providers/curation_api_client.py
chelseatroy: Went ahead with your approach here! Extracted variables with names to make it clear why we're doing this.

app/data_providers/curation_api_client.py
```diff
@@ -0,0 +1,48 @@
...
+        if not start_date:
+            start_date = date.today().strftime("%Y-%m-%d")
+
+        request_headers = {
+            "apollographql-client-name": "recommendations-api",
+            "apollographql-client-version": "1"
+        }
+
+        injection_protected_id = str(corpus_id)
+        injection_protected_start_date = str(start_date)
```
mmiermans, line 29: Converting to string doesn't protect against injection. Try running this:
```suggestion
        start_date = '2022-02-22", arg2: "blah"'
        injection_protected_start_date = str(start_date)
```
Note that the date string contains an unesceaped `"` character. The GraphQL interpreter will interpret that as the end of the string, and treat `arg2` as a second query parameter to `items`. Adding a query parameter is pretty harmless, but maybe a specially crafted string could run a totally different query or mutation.

I'll see if I can get the JSON stringify to work quickly.

app/data_providers/metrics_client.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, firefox_newtab_metrics_factory):
+        self.firefox_newtab_metrics_factory = firefox_newtab_metrics_factory
+
+    async def get_engagement_metrics(self, ranked_items, ranker):
+        ranker_kwargs = {}
+
+        # If this turns into a skyscraper of conditionals I recommend we consider a jump table instead
+        if ranker is firefox_thompson_sampling_1day:
+            ranker_kwargs = {
+                'metrics': self.firefox_newtab_metrics_factory.get([rec.id for rec in ranked_items])
```
mmiermans, line 20: When I ran a query I got an exception:
> RuntimeWarning: coroutine 'FirefoxNewTabMetricsFactory.get' was never awaited
```graphql
query corpus_slate {
  getRankedCorpusSlate(slateId: "f99178fb-6bd0-4fa1-8109-cda181b697f6") {
    corpusItems {
      id
    }
  }
}
```

This should fix it:
```suggestion
                'metrics': await self.firefox_newtab_metrics_factory.get([rec.id for rec in ranked_items])
```

tests/unit/data_providers/test_metrics_client.py
```diff
@@ -0,0 +1,42 @@
...
+            scheduled_surface_item_id='4a105732-6dcc-4bfa-a92e-8bb0e5616e89',
+            slate_experiment_id='13055e0',
+            url='https://example.com/00000000-0000-0000-0000-000000000000',
+            slate_id='f99178fb-6bd0-4fa1-8109-cda181b697f6',
+            trailing_1_day_opens=0,
+            trailing_1_day_impressions=100000
+        )
+    }
+
+    def get(self, recommendation_ids):
```
mmiermans, line 23: This is why the 'coroutine was never awaited' issue didn't happen in tests:
```suggestion
    async def get(self, recommendation_ids):
```

mmiermans (commented):
I created a PR that you can merge into this one: https://github.com/Pocket/recommendation-api/pull/683

mmiermans (approved):
:+1:

PR #678, opened 2022-02-22
Snowplow events

chelseatroy:
# Goal

We have to send a bunch of IDs to Snowplow in order to be able to track recommendation events such that we can map them to the tiles that allow us to track engagements. 

Tickets:
* https://getpocket.atlassian.net/jira/software/projects/FFRECSV2/boards/86?assignee=5fd7e353dd5eb50108f236c8&selectedIssue=FFRECSV2-170

## Implementation Decisions

Mostly commented or indicated in the code, but:

- Same dependency injection situation as the rest of this flow
- We've got some new IDs that we're literally just generating UUIDs for as per a conversation with @gkatre and @jeshuaborges. I checked the spec for Python UUID generation and as of Python 3.7 it is in line with UUID collision likelihood expectations (the number won't mean anything to anyone, but it calculates out to "about once per decade if every person currently on Earth generates a new UUID every second"). 
- Null value for now for the slate experiment id (this is for new tab flows, which we don't experiment on at the moment)


app/data_providers/snowplow_client.py
```diff
@@ -0,0 +1,41 @@
...
+            items: [CorpusItem]=[]
+    ) -> None:
+        tracking_payload = payload.Payload()
+        tracking_payload.add_json(dict_={
+            "recommendation_id": str(uuid.uuid4()),
+            "slate_experiment_id": None,  # A null value for now
+            "user_id": user_id,
+            "items": [
+                {
+                    "recommendation_item_id": str(uuid.uuid4()),
```
jeshuaborges, line 31: How is this id transmitted to the client so that it can be reported with engagement events?

app/data_providers/snowplow_client.py
```diff
@@ -0,0 +1,41 @@
...
+                    "scheduled_corpus_item_id": item.id,
```
gkatre, line 32: To my understanding, the `scheduled_corpus_item_id` is the ID that comes from a UUID created by the scheduling process (instance) of the Corpus-Item. I dont think its the `item_id`

gkatre (commented):
@chelseatroy I left one comment about: `"scheduled_corpus_item_id": item.id`

app/data_providers/snowplow_client.py
chelseatroy: It isn't yet; we get to that in the next ticket: https://getpocket.atlassian.net/jira/software/projects/FFRECSV2/boards/86?selectedIssue=FFRECSV2-169

app/data_providers/snowplow_client.py
chelseatroy: I think it's important to mention here that the reason this is called `item.id` is that I am doing a list comprehension in which the termporary variable is called "item." This is not the canonical "item_id" we have historically had at Pocket. I will change the temporary var name to make that clear, though!

app/data_providers/snowplow_client.py
```diff
@@ -0,0 +1,71 @@
...
+                    "scheduled_corpus_item_id": corpus_item.id,
+                } for corpus_item in items
+            ],
+        },
+            encode_base64=False,
+            type_when_encoded="encoded_recs_api_recommendations",
+            type_when_not_encoded="recs_api_recommendations"
+        )
+
+        self.tracker.track_self_describing_event(
```
chelseatroy, line 43: @bassrock this is where we're sending the event.

jeshuaborges:
This is handled by a new body of work.

PR #816, opened 2022-05-04
Add read access to personalizer-<env>-UserTopicProfileStore

izo3:
# Goal

Allow read access to personalizer-<env>-UserTopicProfileStore dynamodb table from ECS task role. 

Tickets:
* POC-98


mmiermans (changes requested):
I believe this PR is out of date and can be closed?

PR #851, opened 2022-05-25
add per-user impression filtering using feature store

lfishhead:
# Goal

This PR adds a ranker that filters items from a slate if the user has seen the item too many times, or first saw it too long ago.  Items to be filtered are retrieved from the feature store given a `user_id`.   

Our current persoanlized curated slate shows the same items fto the same users for too long.  We would like to increase churn there using this ranker to furthr improve that slate.

## Implementation Decisions

1. we are using `user_id` as elsewhere in the repo, as opposed to the hashed `user_id` here and in the feature group
2. we are using the `resolved_id` as elsewhere in the repo as opposed to the url or other identifiers here and in the feature group
3. This [PR](https://github.com/Pocket/data-flows/pull/50) must be deployed for the feature store to have updated item lists per-user

## References

[confluence](https://getpocket.atlassian.net/l/c/7ShvyL0E)

### DIFF
--- a/.run/main.run.xml
+++ b/.run/main.run.xml
@@ -6,7 +6,7 @@
     <envs>
       <env name="PYTHONUNBUFFERED" value="1" />
     </envs>
-    <option name="SDK_HOME" value="docker-compose://[$PROJECT_DIR$/docker-compose-dev.yml]:app_dev/python" />
+    <option name="SDK_HOME" value="docker-compose://[$PROJECT_DIR$/docker-compose.yml]:app_dev/python" />
     <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$" />
     <option name="IS_MODULE_SDK" value="false" />
     <option name="ADD_CONTENT_ROOTS" value="true" />
--- a/.run/main.run.xml
+++ b/.run/main.run.xml
@@ -6,7 +6,7 @@
     <envs>
       <env name="PYTHONUNBUFFERED" value="1" />
     </envs>
-    <option name="SDK_HOME" value="docker-compose://[$PROJECT_DIR$/docker-compose.yml]:app_dev/python" />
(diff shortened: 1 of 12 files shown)


.run/main.run.xml
```diff
@@ -6,7 +6,7 @@
     <envs>
       <env name="PYTHONUNBUFFERED" value="1" />
     </envs>
-    <option name="SDK_HOME" value="docker-compose://[$PROJECT_DIR$/docker-compose.yml]:app_dev/python" />
+    <option name="SDK_HOME" value="docker-compose://[$PROJECT_DIR$/docker-compose-dev.yml]:app_dev/python" />
```
mmiermans, line 9: Could you revert this change?

I mentioned this live as well, but we didn't have an easy way to test your changes end-to-end against realistic data. Something for me to think about as I'm going back to RecommendationAPI for the Setup Moment project.

tests/unit/rankers/test_algorithms.py
```diff
@@ -318,3 +318,44 @@ def test_no_topic_slates(self):
 
         for topic_ranker in [top1_topics, top3_topics, rank_topics]:
             self.assertRaises(ValueError, topic_ranker, input_configs, full_topic_profile)
+
+
+class TestAlgorithmsImpressionFilter(unittest.TestCase):
```
mmiermans, line 323: :+1: These tests for `user_impression_filter` look rock solid.

app/json/slate_configs.json
```diff
@@ -272,6 +272,7 @@
           "recit-personalized/curated"
         ],
         "rankers": [
+          "user-impression-filter",
```
mmiermans, line 275: Are you planning to monitor the effect on this slate, before applying this ranker more widely?

mmiermans (commented):
I'm supportive of deploying this. We did pair on this extensively, so (if possible) I'd like to get an outside perspective on this. I think Jesh would be the only one potentially qualified.

mmiermans (approved):
lgtm :+1:

PR #859, opened 2022-06-01
Mock GraphQL endpoints for Setup Moment

mmiermans:
# Goal
Unblock frontend development by mocking the queries `setupMomentSlate` and `userContentProfileTopics`, and the mutation `updateUserContentProfile`.

## Todos
- [x] Tested locally
- [x] Tested in Pocket-Dev

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/DPS-5

Spec:
* https://github.com/Pocket/spec/pull/144

## Implementation Decisions
- Focus on the endpoints need for Setup Moment onboarding. Eventually we'll add a query `homeSlateLineup` to power Home using Corpus Items, but neither we nor frontend will work on that until a few weeks from now.
- Punting on tests for now, because I'd like to get this out quickly, and these endpoints are only used for development for now.


gkatre (approved):
@mmiermans and I reviewed this together in a paring session

PR #874, opened 2022-06-06
fix for android lineup

lfishhead:
# Goal

Android request to recs-api is raising personalization error and failing to return any recommendations to android discover.

## Reference

[slack](https://pocket.slack.com/archives/C013MQ3NBN1/p1654545815622619)
## Implementation Decisions

remove raising of personalization error for users without impressed items


mmiermans (approved):
:+1:

PR #884, opened 2022-06-10
[HS-33] Implement updateUserRecommendationPreferences

mmiermans:
# Goal
Implement `updateUserRecommendationPreferences` to store topics that the user has selected in a Feature Group.

## Video walkthrough
https://drive.google.com/file/d/1syaa7XvrXCPwyolF4Hi2wRC3aOQVI1tC/view?usp=sharing

## Todos
- [x] Tested locally

## Deployment
- [x] Created feature group `development-user-recommendation-preferences-v1`
- [x] Created feature group `production-user-recommendation-preferences-v1`

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-33

## Implementation Decisions
- This branch is based off of `fetch-corpus-candidates` because it builds on changes to feature group tests.


app/data_providers/user_recommendation_preferences_provider.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        async with self.aioboto3_session.client('sagemaker-featurestore-runtime') as featurestore:
+            await featurestore.put_record(
+                FeatureGroupName=self.get_feature_group_name(),
+                Record=self._feature_store_record_from_model(model)
+            )
+
+    @classmethod
+    def _feature_store_record_from_model(cls, model: UserRecommendationPreferencesModel) -> List[Dict[str, Any]]:
+        return [
```
jeshuaborges, line 43: I like how you didnt over complicate this by trying to abstract too much serialization or putting in any aggressive libraries.

app/data_providers/user_recommendation_preferences_provider.py
```diff
@@ -0,0 +1,60 @@
...
+from app import config
+from app.models.topic import TopicModel
+from app.models.user_recommendation_preferences import UserRecommendationPreferencesModel
+
+
+class UserRecommendationPreferencesProvider:
+    _FEATURE_GROUP_VERSION = 1
+
+    def __init__(self, aioboto3_session: aioboto3.session.Session = None):
+        self.aioboto3_session = aioboto3_session if aioboto3_session else aioboto3.Session()
```
jeshuaborges, line 16: Why not always inject the session?

app/graphql/update_user_recommendation_preferences_mutation.py
```diff
@@ -12,8 +16,12 @@ class Arguments:
...
-        topics = await TopicModel.get_all()
-        input_topic_ids = {t.id for t in input.preferredTopics}
-        output_topics = [t for t in topics if t.id in input_topic_ids]
+        model = UserRecommendationPreferencesModel(
+            user_id=info.context.get('user_id'),
+            updated_at=datetime.datetime.utcnow(),
+            preferred_topics=await TopicModel.get_topics({t.id for t in input.preferredTopics})
+        )
+
+        await UserRecommendationPreferencesProvider().put(model)
```
jeshuaborges, line 25: I like the use of the data mapper pattern. It may be worth leaning into some of the patterns you are using here and giving them names, or otherwise document architecture, such that its clear the strategies that you are using. We could add an `Entity` or `Model` suffix to models which would make it more clear which systems are talking to each other and that we are always passing around domain objects.

(not necessary to be addressed in this PR)

app/models/topic.py
```diff
@@ -54,6 +54,16 @@ async def get_all() -> ['TopicModel']:
...
         return sorted(list(map(TopicModel.from_dict, response['Items'])), key=lambda topic: topic.slug)
 
+    @staticmethod
+    async def get_topics(topics_ids: Set[str]) -> ['TopicModel']:
+        """
+        Retrieves all topics from dynamo db
+
+        :return: a list of TopicModel objects
+        """
+        all_topics = await TopicModel.get_all()
```
jeshuaborges, line 64: It looks like you switch to Active Record style models here. Why a different strategy than above with user recommendation preferences?

tests/functional/graphql/test_update_user_recommendation_preferences.py
```diff
@@ -45,11 +50,20 @@ def test_update_user_recommendation_preferences(self, mock_client_session_get):
...
                 executor=AsyncioExecutor())
 
             response = executed.get('data').get('updateUserRecommendationPreferences')
             preferred_topics = response['preferredTopics']
 
             assert len(preferred_topics) == 1
             assert preferred_topics[0]['name'] == 'Technology'
+
+            # Assert that the data provider was called with the right data.
+            assert mock_data_provider_put.call_count == 1
```
jeshuaborges, line 63: Usually, in functional and integration tests we are testing the return values of the system and we are less concerned about the internal workings of this system. This is because functional tests cross many boundaries and the internals may change. If it is needed to validate internal behavior this test should be moved to a unit test.

jeshuaborges (approved):
There is a lot I like about this PR. Nothing I have noted in the PR is a risk for immediate deployment and is not blocking merge. I do feel like it's important to decide on a data access/model strategy and decide between active record and data mapper strategies. I prefer data mapper for a lot of reasons but we can pontificate about at another time.

PR #888, opened 2022-06-12
[HS-42] Rank setup moment by user preferred topics

mmiermans:
# Goal
Rank `setupMomentSlate` by user preferred topics.

## Todos
- [ ] Test locally using Postman or cUrl:
```shell
curl --location --request POST 'http://localhost:8000' \
--header 'userId: <user id that exists in preferences feature group>' \
--data-raw '{
    "query": "query PostmanSetupMomentSlate { setupMomentSlate{ id headline recommendations { id corpusItem { id } } }}",
    "variables": null,
    "operationName": "PostmanSetupMomentSlate"
}'
```

## Deployment
- [ ] Merge https://github.com/Pocket/recommendation-api/pull/887
- [ ] Add `corpus_topic_id` to production DynamoDB table `RECAPI-Prod-TopicMetadata`
- [ ] Merge this PR

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-42

## Implementation Decisions
- Keep `SetupMomentDispatch` separate from `RankingDispatch` to punt on generalizing CorpusItem slates. In particular, the Firefox slates are different because they pull corpus items from the CuratedCorpusAPI instead of the Sagemaker Feature Store. Maybe a factory method pattern can help decide which corpus client to instantiate? Or maybe we want to load the scheduled surfaces in a Feature Store as well?

### DIFF
--- a/app/rankers/algorithms.py
+++ b/app/rankers/algorithms.py
@@ -4,8 +4,7 @@ import json
 
 from aws_xray_sdk.core import xray_recorder
 
-from app.data_providers.slate_provider_schemata import ExperimentSchema
-from app.graphql.corpus_item import CorpusItem
+from app.models.corpus_item_model import CorpusItemModel
 from app.models.metrics.metrics_model import MetricsModel
 from app.models.metrics.firefox_new_tab_metrics_model import FirefoxNewTabMetricsModel
 
@@ -16,6 +15,7 @@ from scipy.stats import beta
 
 from app.models.slate_config import SlateConfigModel
 from app.models.personalized_topic_list import PersonalizedTopicList
+from app.models.topic import TopicModel
 
 DEFAULT_ALPHA_PRIOR = 0.02
 DEFAULT_BETA_PRIOR = 1.0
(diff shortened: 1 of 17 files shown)


app/rankers/algorithms.py
```diff
@@ -242,6 +243,19 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
...
+        recs: CorpusRecommendationListType,
+        preferred_topics: List[TopicModel],
+) -> CorpusRecommendationListType:
+    """
+    Sorts recommendations in-order such that recs with a preferred topic rank above recs with a non-preferred topic.
+    :param recs:
+    :param preferred_topics: List of topics that the user has expressed a preference for.
+    :return: ordered list of recommendations with preferred topics ranking higher.
+    """
+    preferred_corpus_topic_ids = {topic.corpus_topic_id for topic in preferred_topics}
```
jeshuaborges, line 256: Nit: This is probably worth a unit test if for no more reason than giving a very specific example.

jeshuaborges (approved):
Looks great to me.

app/rankers/algorithms.py
mmiermans: Good call-out. I've added a ticket to track this under 'cleanup' just so I don't forget. Should only take a minute: https://getpocket.atlassian.net/browse/HS-64

PR #889, opened 2022-06-13
[HS-62] Allow clients to limit recommendation count in setupMomentSlate

mmiermans:
# Goal
Allow clients to limit the number of recommendation that `setupMomentSlate` returns.

## Todos
- [ ] Test locally
- [ ] Test in Apollo using a Pocket-Dev deployment

## Reference

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/HS-62)

## Implementation Decisions
- As far as I can tell, Graphene does not make it easy to access field arguments, so I wrote a utility function for this.

### DIFF
--- a/tests/unit/graphql/test_get_field_argument.py
+++ b/tests/unit/graphql/test_get_field_argument.py
@@ -1,0 +1,83 @@
+import unittest
+
+from graphql.language.ast import Field, Name, SelectionSet, Argument, IntValue
+
+from app.graphql.util import get_field_argument
+
+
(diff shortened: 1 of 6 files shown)


tests/unit/graphql/test_get_field_argument.py
```diff
@@ -0,0 +1,83 @@
...
+                ])
+            )
+        ])
+    )
+
+
+_setup_moment_slate_fields_fixture = [_setup_moment_slate_field_fixture()]
+
+
+class TestGetFieldArgument(unittest.TestCase):
```
jeshuaborges, line 32: Nice work adding these tests.

PR #914, opened 2022-06-29
Emit a CorpusSlate Snowplow event in setupMomentSlate

mmiermans:
# Goal
Send a Snowplow event when `setupMomentSlate` is requested that associates `CorpusRecommendation.id` the metadata defined in https://github.com/Pocket/spec/pull/147. The Web client will only need to emit `CorpusRecommendation.id`, and Data Products will have direct control over the metadata we need to associate with these recommendations.

Changes:
- Change `CorpusRecommendation.id` to the [RFC 4122](https://tools.ietf.org/html/rfc4122) uuid spec, instead of a `hex` format without dashes. In https://github.com/Pocket/spec/pull/147 we've specified that identifiers need to be in uuid format, and Snowplow therefore rejected the hex format. This identifier was not used or stored anywhere yet, so it's safe to change.

# Todo:
- [x] Change schema names from shorthand (e.g. `CorpusSlate.id` to long names `CorpusSlate.corpus_slate_id`.

# Deployment
- [x] Deployed Snowplow entity changes to production

#### Steps:
1. Run `source <(maws -o awscli)` and auth with pocket-dev-PocketSSODataLearning.
2. In PyCharm, select the 'main Pocket-Dev' run configuration and click 'run'.
3. Make a request to setupMomentSlate (this request is also available the Postman Pocket workspace):
```shell
curl --location --request POST 'http://localhost:8000' \
--header 'userId: 12345' \
--data-raw '{
    "query": "query PostmanSetupMomentSlate { setupMomentSlate{ id headline recommendations(count: 10) { id corpusItem { id } } }}",
    "variables": null,
    "operationName": "PostmanSetupMomentSlate"
}'
```

#### Expectations:
- [x] In the PyCharm Debug Console a success message is logged by Snowplow:
   > INFO:snowplow_tracker.emitters:GET request finished with status code: 200
- Open the [Snowplow Mini Kibana](https://com-getpocket-prod1.mini.snplow.net/kibana/app/kibana#/discover?_g=()&_a=(columns:!(_source),index:good,interval:auto,query:(language:lucene,query:''),sort:!(collector_tstamp,desc))) (access can be requested through the analytics team).
   - [x] There should be an event in the 'good' index corresponding to your request.
   - [x] The event should have a 'corpus_slate' context.
   - [x] The event should have a 'user' context with the user id that you passed in.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-44

Snowplow schema spec PR:
* https://github.com/Pocket/spec/pull/147

## Implementation Decisions
- I refactored the official snowplow-tracker library to support asyncio and published it to pypi as [aio-snowplow-tracker](https://pypi.org/project/aio-snowplow-tracker).
- Add `pocket/snowplow-micro` to docker-compose.yml to run tests against Snowplow in isolation.
- The CircleCI `test_app` job now runs in Docker Compose instead of in a pipenv shell. This resolves some network issues in CircleCI between snowplow and the test code, and it brings the test execution in line with how tests are recommended to be run locally.
- I am trying out a new pattern to store configuration. I'm curious to hear what others think.
   1. Instead of using a dict, I'm storing configuration in a class. This has the benefit of allowing for static code analysis by an IDE, for example to catch typos.
   2. Instead of collecting all configuration in a single `config.py` file I'm putting it in the package/directory that owns the configuration.


bassrock:
Hiya! I saw this in your slack channel and just wanted to add a thought on snowplow testing!

> The functional tests in TestSetupMomentSlate covers this change by hitting Snowplow Mini. Generally I consider it best practice to not use external dependencies in tests, but Snowplow Mini is already used by the QA team for end-to-end testing, so this is a supported use-case. Is this a good decision? 

You could use [snowplow micro](https://github.com/snowplow-incubator/snowplow-micro) for this. We have a docker image that pulls the latest from the schemas at https://hub.docker.com/repository/docker/pocket/snowplow-micro. Examples of this usage in tests can be seen in the [curated corpus api](https://github.com/Pocket/curated-corpus-api/blob/a17ce6ef894df4ab1cf04c78f17ae3dd5b417b7d/src/events/snowplow/ScheduledItemSnowplowHandler.integration.ts#L125-L138)

app/data_providers/snowplow_corpus_slate_tracker.py
```diff
@@ -0,0 +1,103 @@
...
+    Implements tracking the recommendation of a CorpusSlate in SnowPlow, with the goal of tracking any metadata that
+    to CorpusRecommendation.id that we are interested in. Clients will only emit the CorpusRecommendation id in events.
+    """
+
+    def __init__(self, tracker: Tracker, snowplow_config: SnowplowConfig):
+        self.tracker = tracker
+        self.snowplow_config = snowplow_config
+
+    @xray_recorder.capture_async('data_providers.SnowplowCorpusSlateTracker.track')
+    async def track(self, corpus_slate: CorpusSlateModel, user_id: Optional[str]):
```
eu9ene, line 44: why user_id is optional? is there a use case when we don't have it? is it also optional in the schema?

eu9ene:
I think blocking the event loop is definitely not a good idea. It kind of kills the whole point of using asyncio. It seems like it shouldn't be too hard to implement a [custom emitter](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/#custom-emitters) that supports asyncio. We can just copy paste the Emmiter class or inherit it and replace the required methods to asycio ones. https://github.com/snowplow/snowplow-python-tracker/blob/master/snowplow_tracker/emitters.py

app/data_providers/snowplow_corpus_slate_tracker.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import List, Any, Dict, Optional
+
+from aws_xray_sdk.core import xray_recorder
+from snowplow_tracker import Tracker, Subject, SelfDescribingJson, Emitter
+
+from app.models.corpus_slate_model import CorpusSlateModel
+from app.config import ENV, ENV_PROD
+
+
+class CorpusSlateTrackable(ABC):
```
eu9ene, line 12: we have only one tracker for now, so probably don't need this abstract class, unless you need it for tests or smth

mmiermans:
> Also sending statistics about recommendations is not the main functionality and should not fail the request if snowplow is unavailable for example.

Snowplow catches all request exceptions, so calls to `track` should never fail: https://github.com/snowplow/snowplow-python-tracker/blob/master/snowplow_tracker/emitters.py#L215-L220
- It would probably be good to log an error in addition to the warning that Snowplow logs, so we'll see it in Unleash.
- It would also be good to add a test case around this behavior.

> So I would imagine the track method to be async fire and forget with catching and logging the exception inside and not awaiting the result in the main recommendations method.

We'd need a background task to do work after the request finishes, and that is not supported with the library versions that we use now. Specifically, we'd need to upgrade FastAPI and switch from Graphene to Strawberry. Would be nice!

eu9ene:
> We'd need a background task to do work after the request finishes, and that is not supported with the library versions that we use now. Specifically, we'd need to upgrade FastAPI and switch from Graphene to Strawberry. Would be nice!

I think we can just implement the regular fire and forget asyncio pattern: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task. The task will be executed on the event loop and since it's just IO and we'll implement an async emitter it won't block the even loop for long.

mmiermans:
> > We'd need a background task to do work after the request finishes, and that is not supported with the library versions that we use now. Specifically, we'd need to upgrade FastAPI and switch from Graphene to Strawberry. Would be nice!
> 
> I think we can just implement the regular fire and forget asyncio pattern: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task. The task will be executed on the event loop and since it's just IO and we'll implement an async emitter it won't block the even loop for long.

Thanks for pointing that out! I've seen asyncio.Task but never used it myself.

Pipfile
```diff
@@ -44,6 +44,7 @@ aiomcache = "==0.6.*"
 # aiocache will automatically use ujson when available. ujson supports decimal types, and json does not.
 ujson = "*"
 httptools = "*"
+aio-snowplow-tracker = "*"
```
eu9ene, line 47: Great job on addressing the asyncio issue. I see that you forked the repo under your personal account. Maybe we should move this fork to pocket or Mozilla github org so that anyone can update the package with the upstream repo if needed?

app/data_providers/dispatch.py
```diff
@@ -40,15 +43,19 @@ async def get_ranked_corpus_slate(self, user_id: str, recommendation_count: int)
...
 
-        return CorpusSlateModel(
+        corpus_slate = CorpusSlateModel(
             id=self.SETUP_MOMENT_CORPUS_CANDIDATE_SET_ID,
             headline=self.DISPLAY_NAME,
             subheadline=self.SUB_HEADLINE,
             recommendations=recommendations,
         )
 
+        await self.slate_tracker.track(corpus_slate, user_id=user_id)
```
eu9ene, line 55: I still think this should be done async and should not block the response since it's not critical to complete the tracking for the main functionality and it affects latency.

Pipfile
mmiermans: Let's see if it becomes a blocker. I used my personal account because I did this work in my spare time. Over the last year, Snowplow has made 3 changes to their Python library, none of which Pocket really needs: https://github.com/snowplow/snowplow-python-tracker/commits/master.

app/data_providers/dispatch.py
eu9ene: I'm less concerned about latency than about additional downstream dependency which can make the service break if unavailable. We should minimize those, especially if it's a tier1 service.

mmiermans: Jira task is tracked here: https://getpocket.atlassian.net/jira/software/c/projects/HS/boards/92/roadmap?selectedIssue=HS-94

eu9ene: It's good that if snowplow is unavailable the method won't fail, but I'm a bit concerned about latency in this case. If, say, there is a retrying mechanism inside the library it might take significantly longer to await if snowplow is down. The general practice of graceful degradation is to make sure all the secondary functionality doesn't hurt the main one and even if something is unavailable the service still can return some results with reasonable latency. I would recommend adding todo here to explore this further.

eu9ene (approved):

PR #917, opened 2022-06-30
Add Data Provider for Unleash assignments

mmiermans:
# Goal
- Get Unleash assignment for content v1 test.
- If user is in content v1 treatment variant, show Pocket Hits and Life Hacks slates.

## Out of scope for this PR
- [HS-316](https://getpocket.atlassian.net/browse/HS-316): Only include en-US in the experiment.
- [HS-347](https://getpocket.atlassian.net/browse/HS-347): Emit a [variant_enroll](https://console.snowplowanalytics.com/cf0fba6b-23b3-49a0-9d79-7ce18b8f9618/data-structures/1945896af520f129eb62b20ceef8cca5be4e3285754b7e51093c089a1ccc7ee9) event

## Deployment
- [ ] Create production `temp.data-products.recommendation-api.home.contentv1` Unleash flag with `control` and `treatment` variants.

## Reference

Tickets:
* Request experiment from Unleash: https://getpocket.atlassian.net/browse/HS-346
* Add Pocket Hits: https://getpocket.atlassian.net/browse/HS-163
* Add Life Hacks: https://getpocket.atlassian.net/browse/HS-164


app/data_providers/unleash_provider.py
```diff
@@ -0,0 +1,91 @@
...
+    We should take care to only use assignments for which the required context is passed in.
+    @see https://github.com/Pocket/feature-flags/tree/main/src/unleashClient/strategy
+    """
+
+    def __init__(self, pocket_graph_client_session: PocketGraphClientSession, unleash_config: UnleashConfig):
+        self.pocket_graph_client_session = pocket_graph_client_session
+        self.unleash_config = unleash_config
+
+    @xray_recorder.capture_async('data_providers.UnleashProvider.get_assignments')
+    async def get_assignments(self, names: List[str], user: User) -> List[UnleashAssignmentModel]:
```
eu9ene, line 38: I assume it will be called from the main methods of the API to determine the user's AB group. When an external service is called from tier1 API usually it's a good idea to use retries and fail on timeout and also provide the fallback. https://github.com/jd/tenacity is a good library for retries that can allow doing this pretty easily.

Failing of getting the AB group should not fail the request, so this method should provide a default alternative if the request is failed (in this case probably just an empty list).

mmiermans: You raise an interesting point here. In the past we've chosen to use the [fail-fast](https://en.wikipedia.org/wiki/Fail-fast) pattern to make it easy to notice what is wrong when a system is failing. Maybe it is time to be more fault-tolerant? Because this would be a shift in design, is it OK if we discuss this in retro, and keep it out-of-scope of this PR? I've already added it to [the agenda](https://docs.google.com/document/d/1RITPzyoXLFGI-eXwp7gicoSX4EBcXD5OodlgtJUW6_c/edit#heading=h.bv7ou16dveny) so we don't lose track.

app/data_providers/dispatch.py
```diff
@@ -137,7 +147,15 @@ async def get_slate_lineup(
         else:
             slates += [self.recommended_reads_slate_provider.get_slate()]
 
-        slates += [self.collection_slate_provider.get_slate()]
+        if contentv1_assignment is not None and contentv1_assignment.variant == 'treatment':
```
eu9ene, line 150: will be able to see somewhere that it didn't go to the experiment branch more often than we expect? maybe we should log what branch was used or add some telemetry?

app/data_providers/unleash_provider.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Returns Unleash assignments with certain assignment names that the user is assigned to.
+        :param names:
+        :param user:
+        :return:
+        """
+
+        # Currently FeatureFlags only has a query for all assignments:
+        # https://github.com/Pocket/feature-flags/blob/main/schema.graphql#L25
+        all_assignments = await self._get_all_assignments(user=user)
```
eu9ene, line 56: I wonder if it's possible to resolve experiments for a user on graphQL level and just pass all of them to recAPI instead of calling the graph here

app/data_providers/dispatch.py
mmiermans: - This PR includes [telemetry for the variant and name](https://github.com/Pocket/recommendation-api/pull/917/files#diff-877dd70fd79464930f75be17e641f0d44fa1d51a17c4d659f39b730f0f7b5905R79-R88) in the same event that emits the slate lineup. This makes it easy to join between slate metadata and the experiment.
- [HS-347](https://getpocket.atlassian.net/browse/HS-347) In a future PR we'll emit a [variant_enroll](https://console.snowplowanalytics.com/cf0fba6b-23b3-49a0-9d79-7ce18b8f9618/data-structures/1945896af520f129eb62b20ceef8cca5be4e3285754b7e51093c089a1ccc7ee9) event that's required by the analytics team. Their experiment dashboard shows the split.

eu9ene: ok, I just want to make sure we'll be able to see it if there's a bug somewhere and contentv1_assignment is None or the variant name is wrong

eu9ene (approved):

PR #926, opened 2022-07-07
Spread topics for setup moment

eu9ene:
# Goal

To have diverse topics based on user choice for setup moment slate 

## Todos
- [x] unit tests pass but I need to figure out how to test this locally
- [x] update functional tests

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/HS-74)

## Implementation Decisions
- no publisher spread so far, but it seems publishers are already pretty diverse on the curated list
- probably not the most efficient or elegant implementation but we're dealing with just 200 items for now, so it should be good enough


jeshuaborges (commented):
Is this code ready for review? It looks like it may not work and the parameters seem out of sync with the function declaration. I stopped before diving in any further.

app/rankers/algorithms.py
```diff
@@ -246,15 +248,40 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
...
+    :param recs: candidates
     :param preferred_topics: List of topics that the user has expressed a preference for.
-    :return: ordered list of recommendations with preferred topics ranking higher.
+    :param count: a number of items to recommend
+    :return: ordered list of recommendations with preferred topics.
     """
-    preferred_corpus_topic_ids = {topic.corpus_topic_id for topic in preferred_topics}
-    return sorted(recs, key=lambda rec: rec.topic in preferred_corpus_topic_ids, reverse=True)
+    topic_ids = {topic.corpus_topic_id for topic in preferred_topics}
+    recs_by_topic = defaultdict(list)
```
jeshuaborges, line 262: Ive never used this implementation `defaultdict(list)`. Its cool to learn something new.

jeshuaborges (approved):
I think this looks great. I had a couple of nits. I will give final review to @mmiermans.

app/data_providers/dispatch.py
```diff
@@ -35,11 +35,10 @@ async def get_ranked_corpus_slate(self, user_id: str, recommendation_count: int)
 
         user_recommendation_preferences = await self.user_recommendation_preferences_provider.fetch(user_id)
         if user_recommendation_preferences:
-            items = rank_by_preferred_topics(items, preferred_topics=user_recommendation_preferences.preferred_topics)
+            items = rank_by_preferred_topics(items, user_recommendation_preferences.preferred_topics, recommendation_count)
         else:
-            logging.info(f'SetupMoment is unpersonalized for user {user_id} because no preferences were found.')
+            logging.warning(f'SetupMoment is unpersonalized for user {user_id} because no preferences were found.')
```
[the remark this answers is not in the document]
eu9ene: I assumed we make a user choose at least one topic, then having 0 topics would indicate that something went wrong. Please correct me if I'm wrong. Maybe this shouldn't matter from the API perspective and we should assume that preferences can be empty.

tests/unit/rankers/test_algorithms.py
```diff
@@ -1,22 +1,106 @@
...
+
+    def test_rank_preferred_topics_5_3(self):
+        topics, recs = self._prepare_recs()
+        user_prefs = [topics[0], topics[2], topics[4]]
+
+        reordered = rank_by_preferred_topics(recs, user_prefs, 5)
+
+        assert len(reordered) == 5
+        assert {r.topic for r in reordered} == {pref.corpus_topic_id for pref in user_prefs}
+        for pref in user_prefs:
```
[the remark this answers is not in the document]
eu9ene: good point, renamed

app/rankers/algorithms.py
```diff
@@ -246,15 +248,40 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
...
         recs: CorpusRecommendationListType,
         preferred_topics: List[TopicModel],
+        count: int
 ) -> CorpusRecommendationListType:
     """
-    Sorts recommendations in-order such that recs with a preferred topic rank above recs with a non-preferred topic.
-    :param recs:
+    Recommends items with preferred topics, rotating the topics. Items for each topic are ordered randomly.
+    In case there's not enough items for the preferred topics, it adds the rest of the candidates in the provided order.
+    :param recs: candidates
```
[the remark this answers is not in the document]
eu9ene: all params are there, I added space

app/data_providers/dispatch.py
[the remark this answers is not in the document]
mmiermans: Users are allowed to click 'skip' without choosing any topics: https://www.figma.com/file/YhPKvRxAyyg3MLduZEkW84/Setup-Moment---Design-Specs?node-id=0%3A1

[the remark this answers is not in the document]
eu9ene: ok, I'll change back to info then

app/rankers/algorithms.py
```diff
@@ -246,15 +248,41 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
...
+    recs_by_topic = defaultdict(list)
+    for rec in recs:
+        if rec.topic in topic_ids:
+            recs_by_topic[rec.topic].append(rec)
+
+    res = []
+    recommended = set()
+    while len(res) < count and len(recs_by_topic) > 0:
+        for topic in topic_ids:
+            if topic not in recs_by_topic:
```
mmiermans, line 272: Optional:
```suggestion
            if not recs_by_topic.get(topic):
```
This removes the need for a `del` statement. I initially thought that `pop` could raise an exception when `topic_recs` was empty, until I realized you handled this with the `del` statement.

app/rankers/algorithms.py
```diff
@@ -246,15 +248,41 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
...
+                continue
+            topic_recs = recs_by_topic[topic]
+            position = random.randint(0, len(topic_recs) - 1) if len(topic_recs) > 0 else 0
+            item = topic_recs.pop(position)
+            res.append(item)
+            recommended.add(item.id)
+            if len(topic_recs) == 0:
+                del recs_by_topic[topic]
```
mmiermans, line 280:
```suggestion
```

app/rankers/algorithms.py
```diff
@@ -246,15 +248,41 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
...
+
+    if len(res) < count:
+        res.extend(r for r in recs if r.id not in recommended)
```
mmiermans, line 283: These recs should also be spread by topic, because otherwise people who choose to skip topic selection might all get 3 recs from the same topic.

Here's a test case:
```python
def test_no_preferred_topics(self):
    topics, recs = self._prepare_recs()
    result = rank_by_preferred_topics(recs, [], count=3)
    assert len({r.topic for r in result}) == 3
```

app/rankers/algorithms.py
```diff
@@ -246,15 +248,41 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
 def rank_by_preferred_topics(
         recs: CorpusRecommendationListType,
         preferred_topics: List[TopicModel],
+        count: int
 ) -> CorpusRecommendationListType:
```
mmiermans, line 252: **Optional:** It took me some time to convince myself that there is no possibility for an infinite loop, and that the function behaves as expected. Not required for this PR, but something to consider: could this function be composed of several logical components, that are more readable because they each perform a simple task? For example:
```python
rank_by_preferred_topics(
    spread_topics(
        randomize(recs)
    ),
    preferred_topics
)
```

1. `randomize` simply returns the input in random order.
2. `spread_topics` maximizes the space between recs of the same topic. It can be implemented very similar to what you've done, without the `random.randint` and `extend`, and by looping until the input is empty.
3. `preferred_topics` would behave the same as the original implementation of this function.

mmiermans (commented):
Thanks for implementing this ticket! I have only one change that I think we should make before deploying this, and a few optional suggestions.

app/rankers/algorithms.py
```diff
@@ -246,15 +248,41 @@ def user_impression_filter(recs: RecommendationListType, user_impressed_list: Li
...
+
+    :param recs: candidates
     :param preferred_topics: List of topics that the user has expressed a preference for.
-    :return: ordered list of recommendations with preferred topics ranking higher.
+    :param count: a number of items to recommend
+    :return: ordered list of recommendations with preferred topics.
     """
-    preferred_corpus_topic_ids = {topic.corpus_topic_id for topic in preferred_topics}
-    return sorted(recs, key=lambda rec: rec.topic in preferred_corpus_topic_ids, reverse=True)
+    topic_ids = {topic.corpus_topic_id for topic in preferred_topics}
```
mmiermans, line 262:
```suggestion
    topic_ids = {topic.corpus_topic_id for topic in preferred_topics}
    if not topic_ids:
        topic_ids = random.shuffle(list({rec.topic for rec in recs}))
```
(The `random.shuffle` is optional, but ensures that not everyone without topics gets the same 3 topics. As far as I know this is not a requirement.)

mmiermans (commented):
This would prevent people without preferred topics from getting recs from the same topic.

app/rankers/algorithms.py
eu9ene: This is a good idea. I didn't consider this use case because I thought we force a user to choose at least one topic (this is what many products do).

eu9ene:
@jeshuaborges @mmiermans if it's a regular case that a user can skip choosing topics, maybe it's better to show just the 3 most popular ones instead of randomization? In code, it would be just passing fixed `List[TopicModel]` instead of user prefs to the ranker.

mmiermans (approved):
:+1: Thanks for jumping in here! Looks great!

PR #940, opened 2022-07-27
Warn not error on missing item/slate dynamodb metrics

jeshuaborges:
# Goal

Stop flooding sentry with events for which there is no action.

Screen Shot 2022-07-27 at 9 31 22 AM

### DIFF
--- a/app/models/metrics/abstract_metrics_factory.py
+++ b/app/models/metrics/abstract_metrics_factory.py
@@ -52,7 +52,7 @@ class AbstractMetricsFactory(ABC):
         metrics = {k.split("/")[0]: self.parse_from_record(v) for k, v in metrics.items() if v is not None}
 
         if not metrics:
-            logging.error(f"No metrics for module {module_id} with keys={keys}")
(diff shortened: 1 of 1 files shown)


mmiermans (approved):
This change was at the back of my mind too. Thanks for acting on it.

mmiermans:
Related Slack threads:
- #qa-setup-moment: https://pocket.slack.com/archives/C03JRQCNSR3/p1656537468237629
- #log-data-product-alerts: https://pocket.slack.com/archives/C02KH5U7G79/p1656608343136779?thread_ts=1656525618.108589&cid=C02KH5U7G79

TL;DR:
- This alert goes off if we receive _no_ clicks/impressions for the entire Slate. The only case in which this should happen if something in the analytics pipeline (including clients) is broken, or when we launch a new Slate.
- Sentry attributes the events to "Pocket web-client", and they started around June 23.
- My guess is that the Web Client requests topic slates to render the 'For You' module, without emitting impression/open events.

PR #942, opened 2022-08-04
[HS-131] Write user preferences keyed on hashed user_id

mmiermans:
# Goal
To allow clients to query a User's recommendation preferences, we want to extend the GraphQL User with a `recommendationPreferences` field. https://github.com/Pocket/spec/pull/159 The GraphQL `User.id` field is the _hashed_ user id, while our feature group currently keys preferences on the _integer_ user id (which is deprecated). To extend the GraphQL User object we need to key these preferences on the _hashed_ user id.

Phases:
1. **(This PR) Write preferences to two Feature Groups, keyed respectively on integer user id and hashed user id. [HS-131](https://getpocket.atlassian.net/browse/HS-131)**
2. Backfill from feature group v1 to v2. [HS-132](https://getpocket.atlassian.net/browse/HS-132)
3. Recommendation API reads from v2 keyed on hashed user_id. [HS-133](https://getpocket.atlassian.net/browse/HS-133)
4. Remove writes to v1 feature group to decommission it. [HS-134](https://getpocket.atlassian.net/browse/HS-134)

## Deployment
- [x] Create `development-user-recommendation-preferences-v2`
- [x] Create `production-user-recommendation-preferences-v2`

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-131

Slack:
- [Thread about whether the integer user id is deprecated](https://pocket.slack.com/archives/C02JZ4TRF0S/p1659979324866529?thread_ts=1659650863.471869&cid=C02JZ4TRF0S)

## Implementation Decisions
- The GraphQL User object and Snowplow user entity have a different meaning for what a "user id" is, which leads to ambiguity in systems that connect to both.
    - My understanding is that using the integer user id is deprecated:
        > My primary take away is that if we are building something new we should use hashed user id and use it as a string key field.
    - I've continued to refer as "hashed_user_id" to the encoded id that GraphQL uses for `User.id`, because this is how we refer to this identifier in practice today.
    - I tried to disambiguate by adding comments.

## Feature Group schema
The schema between v1 and v2 is unchanged, but the meaning of `user_id` has changed from the integer user id to the hashed user id.
```python
[
    {
        # Hashed user id, corresponding in GraphQL to User.id
        "name": "user_id",
        "type": "string"
    },
    {
        "name": "updated_at",
        "type": "datetime"
    },
    {
        # JSON array of topics, e.g. [{"id":"1bf756c0-632f-49e8-9cce-324f38f4cc71"}, {"id":"45f8e740-42e0-4f54-8363-21310a084f1f"}]
        "name": "preferred_topics",
        "type": "string"
    },
]
```

## QA
1. Get your test account user id and hashed/encoded user id.
2. Connect to the Development VPN.
3. Run a query to update preferred topics:
    ```shell
    curl --location --request POST 'https://recommendation-api.getpocket.dev' \
    --header 'userId: put-your-integer-user-id-here' \
    --header 'encodedId: put-your-hashed-user-id-here' \
    --data-raw '{
        "query": "mutation PostmanUpdateUserRecommendationPreferences { updateUserRecommendationPreferences(input: { preferredTopics: [ {id:\"1bf756c0-632f-49e8-9cce-324f38f4cc71\"}, {id:\"45f8e740-42e0-4f54-8363-21310a084f1f\"} ] }) { preferredTopics { id name } }}",
        "variables": null,
        "operationName": "PostmanUpdateUserRecommendationPreferences"
    }'
    ```
4. Query the v1 feature group in Athena:
    ```sql
    SELECT * FROM "sagemaker_featurestore"."development-user-recommendation-preferences-v1-1654826050" limit 10;
    ```
5. Query the v2 feature group in Athena:
    ```sql
    SELECT * FROM "sagemaker_featurestore"."development-user-recommendation-preferences-v2-1659991605" limit 10;
    ```

### QA expectations
- [x] The request successfully returned the two preferred topic ids and names.
- [x] The v1 feature group should have an updated record for the given integer user id
- [x] The v2 feature group should have an updated record for the given hashed user id

### DIFF
--- a/app/data_providers/user_recommendation_preferences_provider.py
+++ b/app/data_providers/user_recommendation_preferences_provider.py
@@ -12,6 +12,10 @@ from app.models.user_recommendation_preferences import UserRecommendationPrefere
 
 
 class UserRecommendationPreferencesProvider:
+    """
+    Put or fetch user preferred topics in a Feature Group keyed on the integer user id.
+    """
+
     _FEATURE_GROUP_VERSION = 1
     _FEATURE_NAMES: List[str] = ['user_id', 'updated_at', 'preferred_topics']
 
@@ -117,4 +121,7 @@ class UserRecommendationPreferencesProvider:
 
 
 class UserRecommendationPreferencesProviderV2(UserRecommendationPreferencesProvider):
+    """
+    Put or fetch user preferred topics in a Feature Group keyed on the hashed user id.
+    """
     _FEATURE_GROUP_VERSION = 2
--- a/app/data_providers/user_recommendation_preferences_provider.py
+++ b/app/data_providers/user_recommendation_preferences_provider.py
@@ -8,7 +8,8 @@ from aws_xray_sdk.core import xray_recorder
 from app import config
 from app.data_providers.topic_provider import TopicProvider
 from app.models.topic import TopicModel
-from app.models.user_recommendation_preferences import UserRecommendationPreferencesModel
+from app.models.user_recommendation_preferences import UserRecommendationPreferencesModel, \
+    UserRecommendationPreferencesModelV2
 
 
 class UserRecommendationPreferencesProvider:
@@ -18,6 +19,7 @@ class UserRecommendationPreferencesProvider:
 
     _FEATURE_GROUP_VERSION = 1
     _FEATURE_NAMES: List[str] = ['user_id', 'updated_at', 'preferred_topics']
+    _FEATURE_ID_NAME: str = 'user_id'
 
     def __init__(self, aioboto3_session: aioboto3.session.Session, topic_provider: TopicProvider):
         self.aioboto3_session = aioboto3_session
@@ -93,7 +95,7 @@ class UserRecommendationPreferencesProvider:
         preferred_topics = json.loads(record['preferred_topics'])
 
         return UserRecommendationPreferencesModel(
-            user_id=record['user_id'],
+            user_id=record[self._FEATURE_ID_NAME],
             updated_at=dateutil.parser.isoparse(record['updated_at']),
             preferred_topics=await self.topic_provider.get_topics([t['id'] for t in preferred_topics])
         )
@@ -102,8 +104,8 @@ class UserRecommendationPreferencesProvider:
     def _feature_store_record_from_model(cls, model: UserRecommendationPreferencesModel) -> List[Dict[str, Any]]:
         return [
             {
-                'FeatureName': 'user_id',
-                'ValueAsString': model.user_id
+                'FeatureName': cls._FEATURE_ID_NAME,
+                'ValueAsString': getattr(model, cls._FEATURE_ID_NAME)
             },
             {
                 'FeatureName': 'updated_at',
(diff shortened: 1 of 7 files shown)


app/data_providers/user_recommendation_preferences_provider.py
```diff
@@ -8,12 +8,18 @@
...
 
 
 class UserRecommendationPreferencesProvider:
+    """
+    Put or fetch user preferred topics in a Feature Group keyed on the integer user id.
+    """
+
     _FEATURE_GROUP_VERSION = 1
     _FEATURE_NAMES: List[str] = ['user_id', 'updated_at', 'preferred_topics']
+    _FEATURE_ID_NAME: str = 'user_id'
```
gkatre, line 22: I see that you have used the [hashed_user_id](https://github.com/Pocket/recommendation-api/pull/942/files#diff-e2ad3ff4166c3a19b996b8d510e6469502daf99882db0dbcd2acdff621fbf0a5R131)  for the _str_ UserID elsewhere. Is there a reason you want to use `user_id` here?

mmiermans: This is the v1 feature store that is keyed on the integer user_id.

app/data_providers/user_recommendation_preferences_provider.py
```diff
@@ -98,8 +104,117 @@ async def _model_from_feature_store_record(
...
+                'ValueAsString': json.dumps([cls._topic_feature_from_topic(t) for t in model.preferred_topics])
+            },
+        ]
+
+    @classmethod
+    def _topic_feature_from_topic(cls, topic: TopicModel) -> Dict[str, str]:
+        return {'id': topic.id}
+
+
+class UserRecommendationPreferencesProviderV2:
```
mmiermans, line 125: This class is duplicated from `UserRecommendationPreferencesProvider`. Originally I had created a derived class, but after deciding to rename `user_id` to `hashed_user_id`, the number of changes between the two classes grew. The v1 feature group will be decommissioned, so this duplicate code won't hand around for long.

app/data_providers/user_recommendation_preferences_provider.py
gkatre: > This is the v1 feature store that is keyed on the integer user_id.

For Feature groups, are we using `str` for all fields (including the integer data types)?

gkatre (approved):
LGTM 👍

app/data_providers/user_recommendation_preferences_provider.py
mmiermans: Yes

mmiermans:
The [UpdateUserRecommendationPreferences operation](https://studio.apollographql.com/graph/pocket-client-api/operations?query=75a3c3270871d56bc6126495222e8e458b3ee3b2&queryName=UpdateUserRecommendationPreferences&range=lastMonth&search=updateUserRecommendationPreferences&variant=current) is not in use anymore, so the risk of deployment here is very low. I will manually hit the mutation in production when deployment has succeeded.

PR #946, opened 2022-08-18
feat(Home): Add homeSlateLineup [HS-157]

mmiermans:
# Goal
Create a demonstration for Unified Home, showing Recommendation API is capable of serving a basic Home experience. The first slate will match the one shown in Setup Moment (curated syndicated recs), followed by random topic slates.

## Out-of-scope
- Being production-ready by having a low latency, or being otherwise reliable or performant
- Ranking based on engagement or personalization
- Impression-based filtering
- De-duplication of the same items across multiple slates
- Emitting Snowplow events for Home

## Changes
- Copy schema from the Spec repo to include the new `homeSlateLineup` Query
- Add `HomeDispatch` to create a Home lineup
- Add `TopicSlateProvider` to create topic slates
- Move the Snowplow tracking of the Setup Moment slate from the `SetupMomentDispatch` to the GraphQL resolver, to only track the slate when it is actually recommended. For Home we will emit a separate event in a future PR.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-157

## Implementation Decisions
- I'm hard coding topic to candidate set ids relationship for demonstration purposes. In the future we might move this data to a database / feature group.


app/data_providers/dispatch.py
```diff
@@ -66,11 +67,56 @@ async def get_ranked_corpus_slate(self, user: UserIds, recommendation_count: int
...
+        self.corpus_client = corpus_client
+        self.user_recommendation_preferences_provider = user_recommendation_preferences_provider
+        self.topic_slate_provider = topic_slate_provider
+
+        self.setup_moment_dispatch = self._create_setup_moment_dispatch()
+
+    async def get_slate_lineup(
+            self, user: UserIds, slate_count: int, recommendation_count: int
+    ) -> CorpusSlateLineupModel:
+        slates = [await self.setup_moment_dispatch.get_ranked_corpus_slate(
```
eu9ene, line 92: we should always use `asyncio.gather` when we can call several async functions at the same time

schema.graphql
```diff
@@ -1,383 +1,390 @@
...
-        "Maximum number of recommendations to return in {Slate.recommendations}, defaults to 10"
-        recommendationCount: Int = 10
-    ): SlateLineup @deprecated(reason: "Please use queries specific to the surface ex. setMomentSlate. If a named query for your surface does not yet exit please reach out to the Data Products team and they will happily provide you with a named query.")
+  """
+  Get stories during Setup Moment onboarding that are personalized with user preferences provided during onboarding.
+  """
+  setupMomentSlate: CorpusSlate
+
+  """
+  Get ranked corpus slates and recommendations to deliver a unified Home experience.
```
eu9ene, line 341: If this API is not production ready, let's mark it as "test" or similar

app/data_providers/dispatch.py
mmiermans: This is a single function call. If clearer, I could split this into two lines:
```suggestion
        slates = []
        slates.append(await self.setup_moment_dispatch.get_ranked_corpus_slate(
```

eu9ene: we have another async function on the next line that potentially does IO operations to retrieve topics

eu9ene (approved):

PR #947, opened 2022-08-18
Remove deprecated queries and deprecated Parser Item queries.

jeshuaborges:
# Goal

Notify client developers to request named corpus items queries when implementing new features.

## Todos
- [x] Mark legacy queries deprecated
- [x] Delete unused deprecated queries and handlers

## Implementation Decisions

We are transitioning to named queries for each surface. Additionally these new queries are backed by Corpus Items instead of Parser Items. Let's deprecate existing legacy queries to ensure that clients are notified that they should migrate to new queries.

This is happening because sometimes clients will create new slate implementations and use existing, getSlate and getSlateLineup, queries rather than contact us for a new use case. This will ensure that they are aware that they should not continue to use the more ambiguous queries anymore.


mmiermans (approved):
Thank you for cleaning up. This was on my mind as well. I can remove a bunch of dead code such as `RankingDispatch` in a follow-up PR.

PR #952, opened 2022-09-13
Home fix baseline unpersonalized

mmiermans:
# Goal
Small fixes related to no-signal Baseline Home:
1. The `CorpusRecommendation.reason` type was incorrect in GraphQL. It should be an object with a `name` and `type` according to the [API spec](https://github.com/Pocket/spec/pull/161). 
2. Topic slates were missing an 'Explore more' link.

## Reference

- [Example Apollo query](https://studio.apollographql.com/graph/pocket-client-api/explorer?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABABISIDKANgIYoIAyAlkgjAA4AUAJHglOURIwdJhCQBhCDFToiASVQBKIsAA6SIkQAW5BNTqMWbdqo1atAZ1r1LZzRa3aENMFWPnHRSzABGz13dWT0c%2BATghERQxJAAlF0txABUCdgQQizgIPmYkAGt7Lwt6AA8UDMcYPCoKogBfWrDBZCiYy04BGRQ5Xn5m4VFxKS6VdQcipjBarQE8dhhLeXo4QqKLKprxteiUKnStopQIdiYoafrzvhpEpFqGrfute7qQABoQADcaPCYaXz3LBgQGMtGoQE0Ii1BpJpKgwXIAGwaF51IA&variant=current)


tests/assets/topics.py
```diff
@@ -24,7 +24,7 @@
     corpus_topic_id='TECHNOLOGY',
     name='Technology',
     display_name='Technology',
-    slug='tech',
+    slug='technology',
```
mmiermans, line 27: I confirmed that the slugs on production are correct.

PR #954, opened 2022-10-04
feat(Home): Add 'Recommended Reads' slate [HS-212]

mmiermans:
# Goal
Add an unpersonalized 'Recommended Reads' slate.

Additional changes:
- rename `for_you_slate_provider.py` for naming consistency.
- add `default=None` to recommendation reason field to be more  explicit and get rid of a warning in the IDE.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-212

Documentation:
* [Content requirements doc](https://docs.google.com/document/d/15gBxnXlsKTgGyjZe72VUc43aK9ZoAArP9NaJDNLIaps/edit#heading=h.mljabq907ibf)
* [Copy requirements](https://getpocket.atlassian.net/browse/HS-227)


app/data_providers/slate_providers/recommended_reads_slate_provider.py
```diff
@@ -0,0 +1,20 @@
+from app.data_providers.corpus.corpus_feature_group_client import CorpusFeatureGroupClient
+from app.models.corpus_recommendation_model import CorpusRecommendationModel
+from app.models.corpus_slate_model import CorpusSlateModel
+
+
+class RecommendedReadsSlateProvider:
```
jeshuaborges, line 6: Nit: There is some boilerplate here causing some of the expressive ness of this class to be lost. As we think about future efforts we will probably want to consider creating a base class or other pattern for corpus feature group client slate models...

jeshuaborges (approved):
LGTM. Love how simple it was to add more content.

app/data_providers/slate_providers/recommended_reads_slate_provider.py
mmiermans: Agreed.

PR #955, opened 2022-10-04
feat(Home): Dedupe recommendations

mmiermans:
# Goal
De-duplicate recommended items across slates on Home, such that users don't see the same recommendation twice.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-202

## Implementation Decisions
- Remove recommendation limit in `TopicSlateProvider`, such that some can be removed during de-duplication, and still meet the requested count.

### DIFF
--- a/app/data_providers/dispatch.py
+++ b/app/data_providers/dispatch.py
@@ -121,10 +121,27 @@ class HomeDispatch:
         ]
 
         return CorpusSlateLineupModel(
-            slates=flatten(list(
-                await gather(*slates)
-            )),
+            slates=self._dedupe_and_limit(
+                flatten(list(
+                    await gather(*slates)
+                )),
+                recommendation_count=recommendation_count,
+            ),
         )
+
+    def _dedupe_and_limit(
+        self, slates: List[CorpusSlateModel], recommendation_count: int
+    ) -> List[CorpusSlateModel]:
+        seen_corpus_ids = set()
+
+        for slate in slates:
+            # Remove recommendations that exist in previous slates, and limit count to recommendation_count.
+            slate.recommendations = \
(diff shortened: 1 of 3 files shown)


app/data_providers/dispatch.py
```diff
@@ -121,11 +121,36 @@ async def get_slate_lineup(
...
-            )),
+            slates=self._dedupe_and_limit(
+                flatten(list(
+                    await gather(*slates)
+                )),
+                recommendation_count=recommendation_count,
+            ),
         )
 
+    def _dedupe_and_limit(
```
jeshuaborges, line 132: Major int: You dont use self in this method. This method would have much less context if it were a classmethod.

app/data_providers/dispatch.py
```diff
@@ -121,11 +121,36 @@ async def get_slate_lineup(
...
+        Deduplicate recommendations across slates, and limit the number of recommendations.
+        It is assumed each individual slate consists of unique items, and this function doesn't look for items occurring
+        multiple times in the same slate.
+        :param slates:
+        :param recommendation_count: The maximum number of recommendations for each slate.
+        :return: Slates with duplicates removed and recommendation_count limit applied.
+        """
+        seen_corpus_ids = set()
+
+        for slate in slates:
```
jeshuaborges, line 145: Another major nit: There is a lot of Law of Demeter stuff going on here. One way that you could drastically simplify some of this logic is that you could push the logic down to other classes. The first, not completely thought through, solution that comes to mind looks something like the following.

```
slate.remove_corpus_items(seen_corpus_ids)
seen_corpus_items |= slate.corpus_item_ids
```

This could make our tests much more simple as almost all your testing logic would be moved to the slate class.

jeshuaborges (approved):
More great work.

PR #958, opened 2022-10-05
feat(Home): homeSlateLineup emits snowplow event [HS-210][HS-287][HS-289]

mmiermans:
# Goal
Emit a `corpus_slate_lineup` entity in the `homeSlateLineup` query.

Additional changes:
- [HS-287](https://getpocket.atlassian.net/browse/HS-287): Fix capitalization of 'Explore more Collections' 
- [HS-289](https://getpocket.atlassian.net/browse/HS-289): Change 'For You' subheadline to 'Recommended for your interests'

## Deployment
- [x] Deploy Snowplow changes from https://github.com/Pocket/spec/pull/164

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-210

## Implementation Decisions
- Create a `SlateProvider` subclass that contains some boilerplate to instantiate a slate.
- Use a factory method pattern to create a `TopicSlateProvider`. This way its constructor can accept a `topic` argument, instead of having to pass this argument around through function calls.


app/data_providers/slate_providers/slate_provider.py
```diff
@@ -0,0 +1,99 @@
...
+from uuid import uuid5, UUID
+
+from app.data_providers.corpus.corpus_feature_group_client import CorpusFeatureGroupClient
+from app.models.corpus_item_model import CorpusItemModel
+from app.models.corpus_recommendation_model import CorpusRecommendationModel
+from app.models.corpus_slate_model import CorpusSlateModel
+from app.models.link import LinkModel
+
+
+class SlateProvider(ABC):
```
jeshuaborges, line 12: This is nice.

app/graphql/resolvers/corpus_slate_lineup_resolvers.py
```diff
@@ -46,14 +48,19 @@ async def resolve_home_slate_lineup(root, info: Info) -> CorpusSlateLineup:
...
-        topic_slate_provider=TopicSlateProvider(corpus_client),
+        topic_slate_providers=TopicSlateProviderFactory(corpus_client),
         collection_slate_provider=CollectionSlateProvider(corpus_client),
     ).get_slate_lineup(
         user=user,
         slate_count=slate_count,
         recommendation_count=recommendation_count,
     )
 
+    slate_lineup_tracker = SnowplowCorpusSlateLineupTracker(
```
jeshuaborges, line 59: For your performance work later this week. It seems like we are constructing a lot of session and provider classes in this resolver. I know that boto3 clients in particular can be expensive to build. I would bet we could get some performance improvements by moving these items to statics or other factories where they are created only once. It would be even better if we could inject all of these providers into this function.

jeshuaborges (approved):
This is great. I see continued improvement and that you are striving a good compromise between focusing on what is important now and waiting for patterns to present themselves.

app/graphql/resolvers/corpus_slate_lineup_resolvers.py
mmiermans: Agreed. That's on my radar as well.

PR #959, opened 2022-10-07
feat(Home): Add homeSlateLineup load test [HS-269]

mmiermans:
# Goal
Add load test for homeSlateLineup.

## Reference

Ticket with load test results:
* https://getpocket.atlassian.net/browse/HS-269

Run test with:
```
nvm use && npm ci && artillery run script.yml
```

## Implementation Decisions
- The GraphQL matches the query that the Web client makes


jeshuaborges (approved):
LGTM. Thanks for adding it to the repo. I am sure it will be handy.

PR #964, opened 2022-10-18
fix(Home): Improve latency [HS-337]

mmiermans:
# Goal
Improve latency of `homeSlateLineup`.

## Changes that improve latency
1. **Caching:** Topic and candidate sets are cached. These are easy targets for caching because they change infrequently and are in the blocking path of every Home request.
2. **Background Snowplow events:** Emit Home Snowplow events in the background instead of blocking the response.
3. ~**Recommendation count:** By default, 6 recommendations are requested instead of 10. This means the corpus-api reference resolver latency should decrease, although it was already pretty low (20-50ms).~
4. **Singleton Data Providers:** Re-use the aioboto3 session across requests.

## Other changes
1. **Instrumentation:** X-Ray now works when running the recommendation-api app locally against Pocket-Dev. I wasn't able to get X-Ray working on ECS for recommendation-api. I'm tracking to fix the integration in ECS in [HS-194](https://getpocket.atlassian.net/browse/HS-194).
2. **Upgrade FastAPI:** Initially emitting Snowplow events in the backgorund did not seem to be working, so I tried to upgrade FastAPI.
3. Remove unused function `TopicProvider.get_topic`.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/HS-337

## Implementation Decisions
- Don't change default recommendations count in this PR, and ask Joel to make the change client-side. ~It makes sense to align the default recommendation count with [designs](https://www.figma.com/file/jawk2lhpNJZZWSJqE3DuNn/Unified-Home-Experiments?node-id=153%3A12627). [Joel agreed](https://pocket.slack.com/archives/C03JSA26QGP/p1665436298581379?thread_ts=1665095371.750609&cid=C03JSA26QGP) and I updated the [Confluence API proposal](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2779938817/Unified+Home). It's ok to make this breaking change now because we know that the Web Client is the only user of `homeSlateLineup`. In the future, we will ask clients to explicitly request a recommendation count if they need fewer or more than the default (6).~
- `asyncio.create_task` vs [FastAPI background events](https://strawberry.rocks/docs/integrations/fastapi#setting-background-tasks). Background events currently have a bug where they get cancelled if the client terminates the request. This should be resolved in Starlette 0.21. It seems the only advantage of background requests is to have access to the request object, which we don't need, so `asyncio.create_task` is better in this case.

### Snowplow background events
- [x] When Snowplow fails to emit an event, the [Sentry error](https://sentry.io/organizations/pocket/issues/3683906511/?environment=development&project=5503693&query=is%3Aunresolved) has request context available.
- [x] Putting `await asyncio.sleep(10)` in the Snowplow track function shows that the response is not blocked by Snowplow.

### XRay
- I didn't get XRay to work in ECS yet. Ignore the timings below, because they're from my PC to US-East1.

#### Expired cache 
Screenshot from 2022-10-19 18-16-35

#### Subsequent request
- Cached request takes significantly shorter.
- It's unclear why the last two slates should take 40ms, 75ms. There's nothing more happening there than a cached lookup and a `random.shuffle`. uvicorn uses uvloop, so the eventloop should be really fast. Not sure if the time is spent in the eventloop, if this is something specific to my machine.

Screenshot from 2022-10-19 18-16-20


tests/functional/graphql/test_home_slate_lineup.py
```diff
@@ -162,5 +167,14 @@ def test_unpersonalized_home_slate_lineup(
             recommendation_counts = [len(slate['recommendations']) for slate in slates]
             assert recommendation_counts == len(slates)*[5]  # Each slates has 5 recs each
 
+            await self.wait_for_snowplow_events()
             all_snowplow_events = self.snowplow_micro.get_event_counts()
-            assert all_snowplow_events == {'total': 1, 'good': 1, 'bad': 0}, self.snowplow_micro.get_last_error()
+            assert all_snowplow_events == {'total': 1, 'good': 1, 'bad': 0}
+
+    async def wait_for_snowplow_events(self, max_wait_time: int = 5):
+        # Locally the request to Snowplow gets handled in 0.01s, but in CircleCI we need 1 second.
```
mmiermans, line 175: I've run-run the tests 10 times on CircleCI to ensure they work reliably.

app/data_providers/corpus/corpus_feature_group_client.py
```diff
@@ -46,6 +47,7 @@ def corpus_item_from_dict(cls, obj: Dict[str, str]) -> CorpusItemModel:
         return CorpusItemModel.parse_obj({k.lower(): v for k, v in obj.items()})
 
     @xray_recorder.capture_async('CorpusFeatureGroupClient._query_corpus_items')
+    @cached(ttl=600)
```
eu9ene, line 50: what's the usual size of a candidate set? do you anticipate a significant memory increase because of caching?

app/data_providers/topic_provider.py
```diff
@@ -14,7 +15,8 @@ class TopicProvider:
     def __init__(self, aioboto3_session: aioboto3.session.Session):
         self.aioboto3_session = aioboto3_session
 
-    @xray_recorder.capture_async('models_topic_get_all')
+    @xray_recorder.capture_async('TopicProvider.get_all')
+    @cached(ttl=60, key='TopicProvider.get_all')
```
eu9ene, line 19: Topics are changed very rarely, right? Can we increase TTL here?

eu9ene:
> It makes sense to align the default recommendation count with [designs](https://www.figma.com/file/jawk2lhpNJZZWSJqE3DuNn/Unified-Home-Experiments?node-id=153%3A12627). [Joel agreed](https://pocket.slack.com/archives/C03JSA26QGP/p1665436298581379?thread_ts=1665095371.750609&cid=C03JSA26QGP) and I updated the [Confluence API proposal](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2779938817/Unified+Home). It's ok to make this breaking change now because we know that the Web Client is the only user of homeSlateLineup. In the future, we will ask clients to explicitly request a recommendation count if they need fewer or more than the default (6).

Sounds pretty bad. Clients definitely should control the recommendation count and the default one shouldn't affect anything.

app/data_providers/corpus/corpus_feature_group_client.py
mmiermans: The max size of a Feature Group record is 350kb and at the moment we have ~20 candidate sets, so the memory usage is negligible. We can switch from memory to ElastiCache/Memcached if that changes.

eu9ene:
@mmiermans some questions that don't block this PR:

XRay looks like a great tool to profile a method call, we should be able to do it on the server though, local is not interesting. Can we deploy to dev and profile there?  

Also, do we leave x-ray working in prod? Does it have any overhead? Or is it like a regular telemetry tool that reports timings rather than a profiler?

app/graphql/recommendation.py
```diff
@@ -12,7 +12,7 @@
     model=RecommendationModel,
     # Currently, the `strawberry.federation.type` cannot be combined with `strawberry.experimental.pydantic.type`,
     # so set the key as a directive.
-    directives=[Key("item { itemId }", UNSET)],
+    directives=[Key(fields="item { itemId }", resolvable=UNSET)],
```
eu9ene, line 15: @mmiermans yes, this thing failed on `docker-compose up` for me, I wonder how CI missed it

mmiermans: I thought it was caused by the FastAPI upgrade, but that wouldn't explain why it happened to you. :man_shrugging:

mmiermans:
> XRay looks like a great tool to profile a method call, we should be able to do it on the server though, local is not interesting. Can we deploy to dev and profile there?

Agreed:

> I wasn't able to get X-Ray working on ECS for recommendation-api. I'm tracking to fix the integration in ECS in [HS-194](https://getpocket.atlassian.net/browse/HS-194).

> Also, do we leave x-ray working in prod? Does it have any overhead? Or is it like a regular telemetry tool that reports timings rather than a profiler?

Good questions.
- In production we define a sampler rule in XRay to sample a certain % of all requests. We can create one that specifically targets recommendation-api to control which % of our requests should be sampled.
- It's unclear to me whether it has overhead. At the end of the `homeSlateLineup` request, it sends an UDP request from the application to the XRay Docker sidecar daemon. During the request it keeps track of the segments in-memory using an async recorder. Folks experienced overhead when they enabled it in our big PHP repo, but not in our typescript repos, as far as I'm aware.
- I think the next step should be to get XRay working again in ECS, because we can't rely on local profiling.

PR #966, opened 2022-10-24
Related recommendations

eu9ene:
# Goal

Added API to recommend related items on syndicated article page:
 1. syndicated only (below the article)
 2. from the same publisher (right rail)

## Todos
- [x] API level tests
- [x] Investigate event loop issues in tests 

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/TDP-48
* https://getpocket.atlassian.net/browse/TDP-47

Proposal: https://getpocket.atlassian.net/wiki/spaces/PE/pages/2824568852/API+for+related+article+recommendations

## Implementation Decisions
- To use only corpus items in recommendations to unify with home and analytics pipeline. This significantly limits the corpus to recommend from and produces only generic recommendations.


app/graphql/resolvers/item2item_resolvers.py
```diff
@@ -0,0 +1,46 @@
...
+
+DEFAULT_RECOMMENDATION_COUNT = 10
+
+
+async def resolve_syndicated(root, info: Info,
+        resolved_id: Annotated[strawberry.ID, argument(
+            description='Article ID')],
+        recommendation_count: Annotated[Optional[int], argument(
+            description='Maximum number of recommendations to return, defaults to 10')] = 10,
+) -> List[strawberry.ID]:
```
mmiermans, line 23: I would return a `CorpusSlate` here:
```suggestion
) -> CorpusSlate:
```
This makes the return type more consistent across recommendation endpoints. The main difference is that this returns a list of `CorpusRecommendation`. Clients need to pass the `id` field of `CorpusRecommendation` in analytics events, and this object contains a `CorpusItem` object, which Apollo is able to decorate.

You could choose to return `List[CorpusRecommendation]`, but the Snowplow events that recommendation-api emits assume you either have a slate or a lineup of slates.

app/graphql/resolvers/item2item_resolvers.py
```diff
@@ -0,0 +1,46 @@
...
+from app.singletons import item2item_recommender
+
+DEFAULT_RECOMMENDATION_COUNT = 10
+
+
+async def resolve_syndicated(root, info: Info,
+        resolved_id: Annotated[strawberry.ID, argument(
+            description='Article ID')],
+        recommendation_count: Annotated[Optional[int], argument(
+            description='Maximum number of recommendations to return, defaults to 10')] = 10,
```
mmiermans, line 22: The `CorpusSlate` already contains a [recomendations count parameter](https://github.com/Pocket/recommendation-api/blob/f5b0da4dd8e7e2c14e95232ccb9f2cf4754263c6/app/graphql/resolvers/corpus_slate_lineup_resolvers.py#L39-L43) on the recommendations field.

app/graphql/resolvers/item2item_resolvers.py
```diff
@@ -0,0 +1,46 @@
...
+    return [strawberry.ID(str(id))
+            for id in await dispatch.syndicated(resolved_id=int(resolved_id),
+                                                count=recommendation_count)]
+
+
+async def resolve_publisher(root, info: Info,
+                            resolved_id: Annotated[strawberry.ID, argument(
+                                description='Article ID')],
+                            domain_id: Annotated[strawberry.ID, argument(
+                                description='Domain ID')],
```
mmiermans, line 36: I have to admit I'm not entirely sure what identifier is best here. Would be good to get some feedback from the clients and backend team.

Pipfile
```diff
@@ -22,7 +22,7 @@ httpx = "*"
 freezegun = "*"
 
 [packages]
-sentry-sdk = "*"
+sentry-sdk = "==1.9.10"
```
mmiermans, line 25: Could you put a comment here mentioning what the conflicting dependency is?

mmiermans (commented):
Overall looks good. I would make some changes to the GraphQL endpoint signature.

app/graphql/resolvers/item2item_resolvers.py
eu9ene: This is a different API, which is not a part of home. How is it related to CorpusSlate?

app/graphql/resolvers/item2item_resolvers.py
eu9ene: I forgot to change this part, it was supposed to return `List[CorpusRecommendation]`. I looked at `CorpusSlate` and it is designed specifically for Home and doesn't fit Related recommendations. It's wrong and confusing to call any recommendations a slate, so let's keep the List[CorpusRecommendation] and communicate to whoever is responsible to change the analytics pipeline to process this.

app/data_providers/dispatch.py
```diff
@@ -7,6 +7,8 @@
 
 from aws_xray_sdk.core import xray_recorder
 
+from app.data_providers.item2item import Item2ItemRecommender
+# from app.data_providers.metrics_client import MetricsFetchable
```
mmiermans, line 11:
```suggestion
```

app/data_providers/dispatch.py
```diff
@@ -25,6 +27,27 @@
...
+    def __init__(self,
+                 # metrics_client: MetricsFetchable,
+                 item_recommender: Item2ItemRecommender):
+        # self.metrics_client = metrics_client
+        self.item_recommender = item_recommender
+
+    async def syndicated(self, resolved_id: int, count: int) -> List[CorpusRecommendationModel]:
+        recs = await self.item_recommender.syndicated(resolved_id, count)
+        # ranked = self.metrics_client.rank_items(recs, rankers=[thompson_sampling_28day])
+        # return ranked
```
mmiermans, line 42: Suggestion: Use a git branch to store future work. Considering it's just a few lines I don't feel super strongly, so up to you.

tests/functional/graphql/test_related.py
```diff
@@ -0,0 +1,203 @@
...
+                    'variables': {
+                        'representations': [
+                            {
+                                '__typename': 'SyndicatedArticle',
+                                'itemId': "111111",
+                                'publisherUrl': "xxxx",
+                            },
+                        ],
+                    },
+                }).json()
```
mmiermans, line 199: Suggestion: add a function to generate this json to make tests easier to read by removing duplicate code.

tests/functional/graphql/test_related.py
```diff
@@ -0,0 +1,203 @@
...
+                            {
+                                '__typename': 'SyndicatedArticle',
+                                'itemId': item_id,
+                                'publisherUrl': pub_url,
+                            },
+                        ],
+                    },
+                }).json()
+
+            print(response)
```
mmiermans, line 76:
```suggestion
```
This is probably debug code while you're fixing the tests.

app/data_providers/dispatch.py
```diff
@@ -25,6 +27,27 @@
...
+# todo: add thompson sampling
+class Item2ItemDispatch:
+
+    def __init__(self,
+                 # metrics_client: MetricsFetchable,
+                 item_recommender: Item2ItemRecommender):
+        # self.metrics_client = metrics_client
+        self.item_recommender = item_recommender
+
+    async def syndicated(self, resolved_id: int, count: int) -> List[CorpusRecommendationModel]:
```
mmiermans, line 39: Nit: I always use verbs for functions, e.g. `recommend_syndicated`, but the code is readable as-is, so I don't feel strongly here.

app/data_providers/item2item.py
```diff
@@ -0,0 +1,82 @@
...
+    return Filter(must=conditions)
+
+
+class Item2ItemRecommender:
+    def __init__(self):
+        host = app.config.qdrant["host"]
+        port = app.config.qdrant["port"]
+        https = app.config.qdrant["https"]
+        self.collection = app.config.qdrant["collection"]
+        self._client = AsyncApis(host=f"http{'s' if https else ''}://{host}:{port}").points_api
```
mmiermans, line 42: Nit: We've started a pattern of injecting dependencies into the constructor to make code more testable and less strongly connected. Any reason not to do that here?

app/data_providers/item2item.py
```diff
@@ -0,0 +1,82 @@
...
+        except UnexpectedResponse as ex:
+            if ex.status_code == 404:
+                # point or collection does not exist
+                # it can happen when a new syndicated article was just added or qdrant state was reset by accident
+                logging.warning(f'Qdrant error: {ex}, returning empty recommendations')
+                return []
+            else:
+                raise
+
+        # print(f'Item2item: recommended: {[rec.payload["title"] for rec in recommended.result]}')
```
mmiermans, line 80:
```suggestion
```

app/graphql/graphql_router.py
```diff
@@ -64,4 +67,4 @@ class Mutation:
     )
 
 
-schema = strawberry.federation.Schema(Query, mutation=Mutation, types=[User], enable_federation_2=True)
+schema = strawberry.federation.Schema(Query, mutation=Mutation, types=[User, SyndicatedArticle], enable_federation_2=True)
```
mmiermans, line 70:
```suggestion
types = [User, SyndicatedArticle]

schema = strawberry.federation.Schema(Query, mutation=Mutation, types=types, enable_federation_2=True)
```
Let's keep lines at 120 characters or less.

tests/functional/graphql/test_related.py
```diff
@@ -0,0 +1,203 @@
...
+                                'itemId': "111111",
+                                'publisherUrl': "xxxx",
+                            },
+                        ],
+                    },
+                }).json()
+
+            assert not response.get('errors')
+            recs = response['data']['_entities'][0]['relatedRightRail']
+            assert len(recs) == 0
```
mmiermans, line 203: Nit: Configure PyCharm to automatically add a newline to the end of files in Editor > General > 'Ensure every saved file ends with a line break'.

mmiermans (approved):
Very readable code. Nice and short! :1st_place_medal: 

I only have optional suggestions.

app/data_providers/item2item.py
eu9ene: I prefer high-level tests, they are less fragile after refactorings. When we want to test this part separately from the connection to qdrant, we'll extract it.

app/data_providers/dispatch.py
eu9ene: long names are not pythonic, I think it's fine to have non-verbs if it's readable

tests/functional/graphql/test_related.py
eu9ene: Sometimes it's ok to duplicate code to have the whole test case inside the test, it's easier to understand tests that way

PR #967, opened 2022-10-24
[HS-281] combine topic and publisher spread

lfishhead:
# Goal

Initial implementation of a ranker that spreads publishers and topics simultaneously.

## Reference

[HS-281](https://getpocket.atlassian.net/browse/HS-281)

## Implementation Decisions

I start by spreading by topics since these may reflect explicit user preferences.  If possible, once we've disqualified some of the available candidates based on their topic, we look for items that allow publisher spreading with a default gap of 2 items.  When this is not possible, we return the next available candidate, i.e. no items are removed from the provide candidate list, only reordered.


lfishhead:
I would guess that the change to the corpus item model will break other tests.  I made publisher a required field there.

tests/unit/rankers/test_algorithms.py
```diff
@@ -24,18 +26,93 @@ class MockCorpusItems:
...
+        assert c10[t] == 2
+        assert c15[t] == 3
+
+
+
+def test_spread_topics_publishers():
+    recs = MockCorpusItems.get_recs_with_publishers()
+    topics = set([r.topic for r in recs])
+
+    for pub_spread in [2, 3, 4]:
```
mmiermans, line 87: Suggestion: A parameterized test makes it a little clearer that the test is executed multiple times and removes a for-loop.
```suggestion
@pytest.mark.parametrize("pub_spread", [(2, ), (3, ), (4, )])
def test_spread_topics_publishers(pub_spread):
    recs = MockCorpusItems.get_recs_with_publishers()
    topics = set([r.topic for r in recs])

```

mmiermans (approved):
Very readable, and you already have tests! :rocket:

app/models/corpus_item_model.py
```diff
@@ -3,4 +3,5 @@
 
 class CorpusItemModel(BaseModel):
     id: str
+    publisher: str
```
mmiermans, line 6: Should this be optional? I'm not sure if we can guarantee to have a publisher for each item? We definitely don't have it backfilled.

lfishhead: and now it is....

lfishhead:
tests are failing since assets lack a publisher field.  `tests/assets/json/sorpus_candidate_sets.json`

mmiermans (commented):
Despite that this is really close, we decided to pick up this PR when we do an epic around ranking improvements, also including Thompson sampling.

Looks like we need to fix a test in this PR. I'd also be curious to do some content QA, and see if there are any differences in practice by this change.

PR #971, opened 2022-10-26
fix(Home): Unclosed client session

mmiermans:
# Goal
Fix ['Unclosed client' error](https://sentry.io/organizations/pocket/issues/3696851013/events/ece585fad96e4e288bd108cfa27822ff/?environment=production&project=5503693&query=is%3Aunresolved&referrer=issue-stream):

> Unclosed client session
> client_session: <app.data_providers.PocketGraphClientSession.PocketGraphClientSession object at 0x7fc18f47d640>

I don't think this has any user impact, but it does eat into our Sentry quota.

### DIFF
--- a/app/graphql/resolvers/corpus_slate_lineup_resolvers.py
+++ b/app/graphql/resolvers/corpus_slate_lineup_resolvers.py
@@ -27,7 +27,6 @@ from app.singletons import (
 
 async def resolve_home_slate_lineup(root, info: Info) -> CorpusSlateLineup:
     user = get_user_ids(info)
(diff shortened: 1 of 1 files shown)


jeshuaborges (approved):
Ah nice. Thanks for handling this. Perhaps there is a cleanup tasks to start reducing the scope of this function in our future.

mmiermans:
> Ah nice. Thanks for handling this. Perhaps there is a cleanup tasks to start reducing the scope of this function in our future.

Yeah, it would be nice to do some automatic dependency injection. That's basically all that this function is doing.

PR #979, opened 2022-11-15
Add fallback to random articles for related

eu9ene:
# Goal

There should always be some recommendations displayed, even if the article was just added to the corpus and vectors store is not updated yet. Fallback to random syndicated or publisher articles.

## Reference

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/TDP-108)

## Implementation Decisions
We could probably come up with something smarter for syndicated, like a filter by `save_count > 10000` but it will require adding another index on the pipeline side, so I don't want to overcomplicate it for now. Unfortunately, qdrant doesn' support ordering for the scroll method

### DIFF
--- a/app/data_providers/item2item.py
+++ b/app/data_providers/item2item.py
@@ -1,3 +1,4 @@
+import logging
 from typing import List
 
 from qdrant_client.http import AsyncApis
@@ -80,6 +81,8 @@ class Item2ItemRecommender:
                         with_vector=False,
                         with_payload=True
                     ))).result.points
+                logging.warning(f'Related: article is not found, fallback to Qdrant scroll method. '
+                                f'resolved_id: {resolved_id}, filter: {query_filter}')
             else:
                 raise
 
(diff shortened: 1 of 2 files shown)


app/data_providers/item2item.py
```diff
@@ -66,16 +66,25 @@ async def _recommend(self, resolved_id: int, query_filter: Filter, count: int):
...
+                # it should fail if the problem is not the article
+                res = (await self._client.scroll_points(
+                    collection_name=self.collection,
+                    scroll_request=ScrollRequest(
+                        limit=count,
+                        filter=query_filter,
+                        with_vector=False,
+                        with_payload=True
+                    ))).result.points
+                logging.warning(f'Related: article is not found, fallback to Qdrant scroll method. '
```
jeshuaborges, line 84: This will be handy.

jeshuaborges (approved):
Looks great.

PR #984, opened 2022-12-22
feat(Home): Add Thompson sampling [TDP-182]

mmiermans:
# Goal
Improve engagement on Home by adding Thompson sampling to the following slates:
- For You
- Recommended Reads
- Popular Collections

## Additional changes
- Cleans up previous Home experiment, which was moved to 100% treatment on Dec 8th

## Review
- [x] Is ranker order correct?
- [x] Are the priors reasonable?

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/TDP-182

## Implementation Decisions
- Continue the current pattern of dependency injection, despite that `corpus_slate_lineup_resolvers.py` has become a bit unwieldy. In a future PR we could adopt a dependency injection library, but I'd prefer to keep the scope of this PR small.
- Batch requests to a Feature Group require some boiler plate code, so I added `FeatureGroupClient` for this purpose. In the future we might choose to also move other Feature Group functionality to this class, but I've left that out-of-scope for this PR.

## Deployment
- [x] Create `temp.web.recommendation-api.home.thompson-sampling` in production

### Thompson sampling is enabled for treatment

Create an Unleash flag in featureflags.getpocket.dev, with 100% going to treatment.

When the `userId` header is a random integer:
- [x] Breakpoints set to the Thompson sampling code block in Recommended Reads and Popular Collections get hit. The engagement metrics variable is populated.
- [x] The recommendation order is not the same on each request.

When the `userId` has preferred topics:
- [x] Breakpoints set to the Thompson sampling code block in For You and Popular Collections get hit.  The engagement metrics variable is populated.
- [x] The recommendation order is not the same on each request.

### Thompson sampling is disabled for control

Set the treatment variant in featureflags.getpocket.dev to 0%.

When the `userId` header is a random integer:
- [x] Breakpoints set to the Thompson sampling code block in Recommended Reads and Popular Collections don't get hit.

When the `userId` has preferred topics:
- [x] Breakpoints set to the Thompson sampling code block in For You and Popular Collections don't get hit.

### Caching
X-Ray unfortunately only works locally at the moment, so the durations in the screenshot below are not representative for production. What it _does_ show is that engagement is cached on the second request.

Screenshot from 2022-12-28 13-40-33


app/data_providers/slate_providers/for_you_slate_provider.py
```diff
@@ -49,7 +62,19 @@ async def rank_corpus_items(
...
+        if kwargs.get('enable_thompson_sampling'):
+            metrics = await self.corpus_engagement_provider.get(
+                self.recommendation_surface_id, self.configuration_id, items)
+
+            items = thompson_sampling(
+                recs=items,
+                metrics=metrics,
+                trailing_period=14,  # A long period might work better given that some topics get few impressions
+                default_alpha_prior=10,  # beta * P95 item CTR for this slate (1.6%)
+                default_beta_prior=600)  # 5% of average daily item impressions for this slate
```
lfishhead, line 74: double parameters here to account for somewhat longer exploration period until prior gets absorbed in clickdata

app/data_providers/slate_providers/recommended_reads_slate_provider.py
```diff
@@ -14,3 +31,26 @@ def headline(self) -> str:
...
+        if kwargs.get('enable_thompson_sampling'):
+            metrics = await self.corpus_engagement_provider.get(
+                self.recommendation_surface_id, self.configuration_id, items)
+
+            items = thompson_sampling(
+                recs=items,
+                metrics=metrics,
+                trailing_period=7,  # With few new items/day and relatively many impressions, a low period is sufficient
+                default_alpha_prior=7,   # beta * P95 item CTR for this slate (0.7%)
+                default_beta_prior=850)  # 5% of average daily item impressions for this slate
```
lfishhead, line 54: i would not mind impression capping here.  if slate turns over once a day, I think there is some risk of items not circulating

lfishhead (approved):
thanks Mathijs, this looks great

app/data_providers/dispatch.py
```diff
@@ -148,29 +148,31 @@ async def get_slate_lineup(
         """
         slates = []
 
-        user_impression_capped_list, preferred_topics, contentv1_assignment = await gather(
+        user_impression_capped_list, preferred_topics, thompson_sampling_assignment = await gather(
             self.user_impression_cap_provider.get(user),
             self._get_preferred_topics(user),
-            self.unleash_provider.get_assignment('temp.web.recommendation-api.home.contentv1', user=user),
+            self.unleash_provider.get_assignment('temp.web.recommendation-api.home.thompson-sampling', user=user),
```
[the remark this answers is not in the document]
mmiermans: I mentioned live that Unleash has some benefits, so just for future reference:
- We have the ability to create experiments in Unleash without being dependent on another team.
- By using Unleash we hook into the analytics ecosystem, which allows us to measure impact on Pocket metrics, such as retention.
- Unleash can easily make experiments sticky to the user id. Less code we need to write in recommendation-api.

That said, we can easily switch to another experimentation framework if needed because our code is modular.

app/data_providers/slate_providers/recommended_reads_slate_provider.py
mmiermans: Created [TDP-228](https://getpocket.atlassian.net/browse/TDP-228) to track.

PR #985, opened 2023-01-05
docs(graphql): don't wrap type names in braces

marcin-kozinski:


mmiermans (commented):
Context:
> Just write your references in plain text and let the tools figure it out. Specifically, don't try wrapping in {}. It doesn't work.

https://pocket.slack.com/archives/C01VDFM30J2/p1672919064066289?thread_ts=1665673637.896369&cid=C01VDFM30J2

mmiermans (approved):

PR #986, opened 2023-01-06
feat(Home): Return German UI strings [TDP-222]

mmiermans:
# Goal
- Add a `locale` argument to `homeSlateLineup`
- If `locale` is "de-DE" then return headlines and subheadlines in German

## Out of scope
- Translating topic slates and topics recommendation reason. [TDP-234](https://getpocket.atlassian.net/browse/TDP-234)
- Switch to German recommendations. [TDP-227](https://getpocket.atlassian.net/browse/TDP-227)
- The Web client will put the German Home behind a feature flag, so no users will see it yet. This allows us to build German Home in smaller PRs and bring those to production to support Web team development.
- Use and document Smartling for translation. [TDP-226](https://getpocket.atlassian.net/browse/TDP-226)

## Reference

- [Jira ticket](https://getpocket.atlassian.net/browse/TDP-222)
- [PRD](https://getpocket.atlassian.net/wiki/spaces/CP/pages/2809135186/PRD%3A+German+Unified+Home+V.1)
- [API proposal](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2866675713/German+Home)

## Review
- Is code readable?
- Do the OOP patterns make sense, or can we do something simpler?

## Deployment
- [ ] Re-autherize CircleCI

## Implementation Decisions
- Move common slate provider dependencies such as `CorpusEngagementProvider` to the parent class `SlateProvider`.
- Use localization for _Pocket Hits_ and _Life Hacks_ slates for consistency, even though they aren't available on German Home as a product requirement.
- Loading the JSON translations file from disk is done synchronously, because it happens only once on the first request and is then cached.

[TDP-234]: https://getpocket.atlassian.net/browse/TDP-234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TDP-227]: https://getpocket.atlassian.net/browse/TDP-227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TDP-226]: https://getpocket.atlassian.net/browse/TDP-226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eu9ene:
I understand that for simplicity we can host translations in this repo, but ideally, they should be in a separate DB managed by the folks responsible for translations. Not sure what's the localization approach at Pocket, but Mozilla uses a community-driven platform https://pontoon.mozilla.org/. Ideally, our solution should be unified with either of them

app/data_providers/dispatch.py
```diff
@@ -105,6 +106,14 @@ class HomeDispatch:
         '45f8e740-42e0-4f54-8363-21310a084f1f',  # Self-improvement
     ]
 
+    # German Home shows different topics by default
+    GERMAN_HOME_TOPICS = [
```
eu9ene, line 110: there might be a better place for those hardcoded values, like some config

app/data_providers/translation.py
```diff
@@ -0,0 +1,59 @@
...
+        return self._strings[item]
+
+    def __contains__(self, item):
+        return item in self._strings
+
+    def get(self, key: str, default: Any):
+        return self._strings.get(key, default)
+
+
+class HomeTranslations(Translations):
```
eu9ene, line 56: why do you need all this "OOP"? it brings extra complexity. I would minimize the abstractions and just simply load the values from the disk to a dict in one function. Smth like dict['locale']['label'].

app/data_providers/translation.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Dict, Any
+
+from app.models.localemodel import LocaleModel
+
+
+class TranslationProvider:
+    def __init__(self, translations_dir: str):
+        self.translations_dir = translations_dir
+
+    @lru_cache(maxsize=None)
```
eu9ene, line 14: lru cache?? do you load many translations in real-time, so that they don't fit in memory and you need to keep only the most recently used ones?  it seems not

app/graphql/locale.py
```diff
@@ -0,0 +1,8 @@
+from typing import NewType
+
+import strawberry
+
+Locale = strawberry.scalar(
```
eu9ene, line 5: do we need this one? can we just use str instead?

app/data_providers/slate_providers/topic_slate_provider.py
```diff
@@ -49,7 +49,13 @@ def headline(self) -> str:
 
     @property
     def more_link(self) -> Optional[LinkModel]:
-        return LinkModel(text=f'Explore more {self.topic.name}', url=f'https://getpocket.com/explore/{self.topic.slug}')
+        # Topic pages only exist for en-US.
+        if self.locale == LocaleModel.en_US:
+            return LinkModel(
+                text=f'Explore more {self.topic.name}',
```
eu9ene, line 55: I see that there's no support for German, but anyway I think all strings should be in one place for consistency, meaning json files.

eu9ene (commented):
I think this service is handling more and more functionality and we should minimize complexity as much as we can. Specifically, get rid of OOP when it's not necessary.

What bothers me more though is that we're adding another responsibility to this service - handling the localization of texts. I don't really believe we're going to AB experiment a lot with phrasing like how we name slates and even if so that this service is the right place for this. I would rather return string templates or IDs here and let clients or apollo pull localizations from other places. I'm curious about what approach is used in other services.

Another concern is that we're inventing a primitive localization platform here, but we should use existing ones (see my other comment).

app/data_providers/translation.py
mmiermans: `@cache` was added in Python 3.9. My understanding is that `@lru_cache(maxsize=None)` is the way to achieve a simple unbounded cache in Python 3.8. I'll add a comment to that effect.

app/graphql/locale.py
mmiermans: My understanding is that this is a way to future-proof this field, by preventing clients from assuming this is a string. In the future we might want to change the type to an Enum. We do the same for [Url](https://studio.apollographql.com/graph/pocket-client-api/schema/reference/scalars?query=Url&variant=current).

I'll ask advice from the backend team whether this is best practice.

mmiermans:
> I understand that for simplicity we can host translations in this repo, but ideally, they should be in a separate DB managed by the folks responsible for translations. Not sure what's the localization approach at Pocket, but Mozilla uses a community-driven platform https://pontoon.mozilla.org/. Ideally, our solution should be unified with either of them

Agreed we should follow the conventional approach. I [double-checked last week](https://pocket.slack.com/archives/C03JSA26QGP/p1672769288532339) that Smartling is still our best-practice at Pocket. I'm keeping this out-of-scope from this PR because I don't want to be blocked on Smartling integration. [TDP-226](https://getpocket.atlassian.net/browse/TDP-226)

app/data_providers/slate_providers/topic_slate_provider.py
mmiermans: Agreed. I kept this out-of-scope because I prefer smaller PRs.

Translating topics would partly follow a similar approach to other headlines, but would require injecting translations into the topics model.

app/data_providers/translation.py
eu9ene: I just don't understand why you need any cache here at all

app/graphql/locale.py
mmiermans: I looked for some more examples and found `CollectionsFiltersInput`, which was recently added and uses `language: String`. I removed the Locale scalar based on that.

app/data_providers/translation.py
```diff
@@ -0,0 +1,21 @@
...
+
+from app.models.localemodel import LocaleModel
+
+
+class TranslationProvider:
+    def __init__(self, translations_dir: str):
+        self.translations_dir = translations_dir
+
+    @lru_cache(maxsize=None)  # Python 3.9 adds a @cache decorator for a simple unbounded function cache.
+    def get_translations(self, locale: LocaleModel, filename: str) -> Dict[str, str]:
```
eu9ene, line 14: why not just load it once on startup and handle extra functionality when it's needed?

eu9ene (approved):

PR #987, opened 2023-01-11
Add the rest of Related API

eu9ene:
# Goal

Add Related API for "after article" and "after save" use cases.

## Todos
- [x] Pydantic conversion + federation doesn't work

## Out of scope
- proper monitoring of the fallback/errors from qdrant

## Reference

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/TDP-121)
* [RFC](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2847637505/RFC+Related+articles+GraphQl+Update)

## Implementation Decisions
- It's better to not fail the request even if qdrant is unavailable which can happen because it is not currently HA
- Fallback to articles with >1000 saves for all requests except publisher based


tests/functional/graphql/test_related.py
```diff
@@ -93,13 +94,107 @@ def publisher_json(item_id: str, pub_url: str):
...
+
+            assert not response.get('errors')
+            entity = response['data']['_entities'][0]
+            recs = entity['relatedAfterCreate']
+            assert entity['itemId'] == item_id
+            assert len(recs) == 3
+            assert 'id' in recs[0]
+            assert 'corpusItem' in recs[0]
+            assert 'id' in recs[0]['corpusItem']
+            print(recs)
```
mmiermans, line 175:
```suggestion
```

tests/functional/graphql/test_related.py
```diff
@@ -93,13 +94,107 @@ def publisher_json(item_id: str, pub_url: str):
...
+            assert not response.get('errors')
+            entity = response['data']['_entities'][0]
+            recs = entity['relatedAfterArticle']
+            assert entity['itemId'] == item_id
+            assert len(recs) == 3
+            assert 'id' in recs[0]
+            assert 'corpusItem' in recs[0]
+            assert 'id' in recs[0]['corpusItem']
+            print(recs)
+            assert all(self.art_by_corpus_id[r['corpusItem']['id']]['is_curated'] for r in recs)
```
mmiermans, line 194: Is the `print` just debug code? If helpful you can optionally add a message to the end of an assertion.
```suggestion
            assert all(self.art_by_corpus_id[r['corpusItem']['id']]['is_curated'] for r in recs), recs
```

app/data_providers/item2item.py
```diff
@@ -68,23 +73,38 @@ async def _recommend(self, resolved_id: int, query_filter: Filter, count: int):
...
+
+        try:
+            res = (await self._client.scroll_points(
+                collection_name=self.collection,
+                scroll_request=ScrollRequest(
+                    limit=count,
+                    filter=query_filter,
+                    with_vector=False,
+                    with_payload=True
+                ))).result.points
```
mmiermans, line 105: This fallback is reasonable for Syndicated and Publisher-based queries, but for the `related` query it could return old curated items that aren't relevant anymore. Would it be better not to return anything in that case?

Usually we QA recommendation results, but I don't see a ticket in [TDP-87](https://getpocket.atlassian.net/browse/TDP-87) to do that. Would you like to create a ticket for that?

mmiermans (approved):
This PR looks good to me, and I only have some optional suggestions. :+1:

app/data_providers/item2item.py
eu9ene: Good question, the assumption is that it's better to show something than nothing, especially if qdrant is down or the article is absent in the vector store which can happen frequently. I will create a ticket for QA.

PR #988, opened 2023-01-12
feat(Home): German recs [TDP-227]

mmiermans:
# Goal
Return localized recommendations for German Home.

## Review
- I'm on the fence on whether having hardcoded candidate sets ids is putting us at risk or holding us back now, or whether it's something we should solve when we want to scale to other markets.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/TDP-227

PR to generate German candidate sets:
* https://github.com/Pocket/data-flows/pull/193

## Implementation Decisions
- Candidate set ids are in a separate file behind a function, to be able to change the implementation in the future without having to make changes to callers.


app/data_providers/slate_providers/candidate_sets.py
```diff
@@ -0,0 +1,61 @@
...
+        'SCIENCE': '785730ff-442d-4d2b-881a-64b01e97585c',
+        'SELF_IMPROVEMENT': '8d2da3af-0ba2-4b9b-836f-a878e081ea5d',
+        'SPORTS': '3d427dc4-e741-41e1-b577-710d2bfa584f',
+        'TECHNOLOGY': '9951fc34-4f66-4838-88f2-b4e0435f6e29',
+        'TRAVEL': '280c14b0-4afb-49d0-b7a5-7a1b156b0c7d',
+    }
+}
+
+
+def get_topic_candidate_set(corpus_topic_id: str, locale: LocaleModel) -> str:
```
lfishhead, line 53: maybe `get_candidate_set_id`?

lfishhead (approved):
looks good to me.

PR #989, opened 2023-01-13
feat(Home): German topic UI [TDP-234]

mmiermans:
# Goal
Provide German localized UI for topic slates.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/TDP-234

## Implementation Decisions
- Delete `setupMomentSlate` to avoid having to make changes in unused code. [Approved by Web team (Joel)](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2877423708/Remove+setupMomentSlate).
- Integrate localization into the `TopicProvider` such that topics are localized by default, as opposed to making this a separate step.
- For consistency, also fetch strings that only exists in English from the translation provider.


app/translations/en-US/topics.json
```diff
@@ -0,0 +1,18 @@
+{
```
eu9ene, line 1: what if we add a new topic? or 100? are all places where we hardcode them documented somewhere? Will migration to a proper translation system fix this?

mmiermans: I don't like how identifiers are hard-coded in several places, in particular UUIDs that aren't human-readable. I started a list here: https://getpocket.atlassian.net/wiki/spaces/PE/pages/2882011170/Topic+References Does that address your concern? If not, could you be more explicit about what you'd like to see changed? If you have any ideas for improvement please let me know. If it's out-of-scope for this PR we can always capture them in tickets.

Integrating with Smartling is out-of-scope for this PR and covered in [TDP-226](https://getpocket.atlassian.net/browse/TDP-226).

mmiermans: The quarterly technology meeting would be a good opportunity to bring up the lack of a service to manage topics. That would likely be a requirement before scaling to 100 topics.

eu9ene: the idea is simple - put them in a db. documenting hardcoding places is a good start

eu9ene (approved):

PR #991, opened 2023-01-17
Change fallback logic for related recs

eu9ene:
# Goal

Make sure we handle fallback properly for different use cases

## Implementation Decisions

- Do not fallback for Similar Stories, return empty
- Fallback to frequently saved or publisher based for other cases
- Cache fallback recommendations to reduce unnecessary calls and return them in case of Qdrant outage (it's not currently HA)
- Request 100 articles on fallback and sample (we can also use thompson sampling later). Randomization can help to diversify content when an article doesn't exist (this will be a frequent case) and gather statistics for different items to use in thompson sampling.
- Never fail request on error from Qdrant because of the secondary nature of "Related" functionality and how it is integrated in the graph (items should always be returned). We'll set up monitoring and alerts based on logs to make sure we can see when errors happen.

[Jira task](https://getpocket.atlassian.net/browse/TDP-267)


app/data_providers/dispatch.py
```diff
@@ -29,22 +30,60 @@
...
+            recs = []
+        return self._sample(recs, count)
+
+    @_empty_on_error
+    async def after_article(self, resolved_id: int, count: int) -> List[CorpusRecommendationModel]:
+        try:
+            recs = await self.item_recommender.related(resolved_id, count)
+        except Item2ItemError:
+            # fallback to frequently saved for "You Might Also Like"
+            recs = await self.item_recommender.frequently_saved(count=100)
```
mmiermans, line 62: I think it's important to make clear in this function's naming that it only returns curated items, for example:
```suggestion
            recs = await self.item_recommender.curated_frequently_saved(count=100)
```

mmiermans (approved):
:+1: lgtm

app/data_providers/dispatch.py
eu9ene: All recommendations provided by this class are currently curated. We are just not allowed to show non-curated stuff. Also, there is `is_syndicated` flag that can make this both curated and syndicated. So I would keep as is.

PR #992, opened 2023-01-18
Standardize logging of related recs

eu9ene:
# Goal

Add custom monitoring of how many errors and fallbacks we have in new related recommendations.

## Reference

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/TDP-123)

## Implementation Decisions

I'm going to use logging parsing like it was done in Recit as the simplest solution. 
The proper approach of course would be to implement either [ structured logging](https://docs.aws.amazon.com/lambda/latest/operatorguide/parse-logs.html) or use a dedicated library for reporting custom application metrics.

### DIFF
--- a/tests/functional/graphql/test_related.py
+++ b/tests/functional/graphql/test_related.py
@@ -4,7 +4,9 @@ import os
 from time import sleep
 from unittest import TestCase, mock
 from unittest.mock import patch
+import re
 
+import pytest
 from fastapi.testclient import TestClient
 from qdrant_client.http.api.points_api import AsyncPointsApi
 from qdrant_client.http.exceptions import UnexpectedResponse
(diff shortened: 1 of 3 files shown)


tests/functional/graphql/test_related.py
```diff
@@ -157,6 +158,8 @@ def after_save_json(item_id: str):
 qdrant_error_mock = mock.Mock()
 qdrant_error_mock.side_effect = UnexpectedResponse(500, 'error', None, None)
 
+logging.getLogger().setLevel(logging.INFO)
```
mmiermans, line 161: Suggestion: Assert that the log statement matches the expected format. For example: https://github.com/Pocket/recommendation-api/blob/main/tests/functional/data_providers/test_snowplow_corpus_slate_tracker.py

mmiermans (approved):
:+1: I like that the solution is simple and not over-engineered. If this becomes a common pattern (which I don't expect), we leverage a [LoggerAdapter](https://docs.python.org/3.10/library/logging.html#loggeradapter-objects) to pass metadata to the logger.

tests/functional/graphql/test_related.py
eu9ene: good idea, this is pretty important for monitoring. added

PR #993, opened 2023-01-19
fix(Home): All users who prefer German should receive German Home [TDP-270]

mmiermans:
# Goal
All users who prefer the German language should receive German Home, including locales such as `de-AT` (Austria).

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/TDP-270

## Implementation Decisions
- [corpus_slate_lineup_resolvers.py](https://github.com/Pocket/recommendation-api/blob/main/app/graphql/resolvers/corpus_slate_lineup_resolvers.py#L34) sets the default using `LocaleModel.from_string(locale, default=LocaleModel.en_US)`. `LocaleModel.from_string` now returns None if the `default` argument is not provided and the locale cannot be matched.

### DIFF
--- a/app/models/localemodel.py
+++ b/app/models/localemodel.py
@@ -16,9 +16,14 @@ class LocaleModel(Enum):
     de_DE = LocaleValue('de-DE')
 
     @classmethod
-    def from_string(cls, val: str, default: en_US) -> 'LocaleModel':
+    def from_string(cls, val: str, default: 'LocaleModel' = None) -> 'LocaleModel':
         for locale in cls:
             if locale.value.lower() == val.lower():
+                return locale
+
+        language = val.split('-')[0].lower()
+        for locale in cls:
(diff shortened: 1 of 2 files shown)


eu9ene (commented):
This probably solves the issue for German, but are we sure that the logic will be the same for all languages? Like we want to show the US edition by default for all English-speaking locales and Spanish(Spain) for the Spanish-speaking world?

mmiermans:
> This probably solves the issue for German, but are we sure that the logic will be the same for all languages? Like we want to show the US edition by default for all English-speaking locales and Spanish(Spain) for the Spanish-speaking world?

For now, yes, everyone except German language users should receive the en-US Home, including GB. But there is not clarity on future market requirements. Quite possibly we'll have to change this logic when those become clear later this year.

eu9ene (approved):

PR #995, opened 2023-01-20
feat(Home): Translations are downloaded from Smartling [TDP-226]

mmiermans:
# Goal
Integrate with Smartling, such that we don't have to manually update individual translations.

## Deployment
- [ ] Give the Data Products team access to Smartling
- [x] Give Julius access to Smartling? Do we want to keep providing translations ourselves?

## Reference

Tickets:
* [TDP-226](https://getpocket.atlassian.net/browse/TDP-226?focusedCommentId=69343)

## Implementation Decisions
- I confirmed that Smartling is [the preferred translation platform at Pocket](https://pocket.slack.com/archives/C03JSA26QGP/p1672769288532339).
- I looked into using the Smartling API to fetch translation strings, but didn't pursue this because the implementation effort doesn't seem to outweigh the effort of downloading and committing translations from Smartling. I documented potential workflow improvements in the readme. More details about my exploration of the Smartling SDK are in [TDP-226](https://getpocket.atlassian.net/browse/TDP-226?focusedCommentId=69343).
- Translations have been coming from our German editor so far. We've only been provided with strings that German Home uses, which does not include strings such as 'For You'. Smartling by default substitutes English translations when no published translations are available in the German locale. Is that confusing or risking that we show German users English by accident? Should we just get everything translated for consistency, even if those strings will not be shown?

[TDP-226]: https://getpocket.atlassian.net/browse/TDP-226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TDP-226]: https://getpocket.atlassian.net/browse/TDP-226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eu9ene:
I think eventually it should be done in the same way as for Firefox translations or pocket web. Example: a [separate repo](https://github.com/mozilla-l10n/firefox-translations-l10n) with strings connected to a translation provider (here pontoon). This repo is added as a submodule to the main git repo. I see that you posted a link to [this one](https://github.com/Pocket/web-localization), which also does this correctly. Why don't we want to do it properly for rec API right away? It seems like it's simply copy-pasting web-localization approach. How many languages are we targeting in near future?

mmiermans:
> Why don't we want to do it properly for rec API right away? It seems like it's simply copy-pasting web-localization approach.

Pocket/web-localization publishes an NPM package that can be shared among multiple Node frontend clients. It's a good approach. I haven't taken it yet, because I don't think it's worth the effort of publishing a Python library when strings are used in a single repo.

> How many languages are we targeting in near future?

No other languages are planned for this half, as far as I'm aware: https://getpocket.atlassian.net/wiki/spaces/CP/pages/2840232077/Discovery+Initiative

eu9ene:
Adding a git submodule looks way simpler than creating a package. Anyway, if no other languages are planned, it's fine to do it manually for now.

mmiermans:
> it's fine to do it manually for now.

:+1: Let's keep other approaches in mind as more localization requirements come up this year. I want to add that both the library and submodule approach require all the same manual steps as listed here, but the advantage is that translations can easily be shared among multiple repos. The main way I see to make the Smartling workflow less manual is to use its API. I think that's totally feasible to do if the current approach becomes burdensome.

PR #1003, opened 2023-01-27
Related diversify publishers

eu9ene:
# Goal

- Add the diversification of publishers for after-article recommendations on the syndicated page.
- Update to v2 qdrant collection requires running the pipeline from https://github.com/Pocket/dl-metaflow-jobs/pull/166 first
- Make sure syndicated items are not shown on the right rail of the syndicated page

# Todo

- [x] merge the metaflow pipeline and run it to get the collection with new domains

# Links
- [Jira ticket](https://getpocket.atlassian.net/browse/TDP-295)


app/data_providers/dispatch.py
```diff
@@ -51,25 +52,29 @@ async def after_save(self, resolved_id: int, count: int) -> List[CorpusRecommend
...
             # do not fallback for "Similar stores" after saving
             recs = []
-        return self._sample(recs, count)
+        return self._to_corpus_items(recs, count)
 
     @_empty_on_error
     async def after_article(self, resolved_id: int, count: int) -> List[CorpusRecommendationModel]:
         try:
-            recs = await self.item_recommender.related(resolved_id, count)
+            recs = await self.item_recommender.related(resolved_id, 20)
```
eu9ene, line 61: it is not guaranteed that there will be 5 different domains in those 20 items, but let's start with this and increase it later if we still see a lot of duplicates after testing

app/data_providers/dispatch.py
```diff
@@ -78,13 +85,26 @@ async def by_publisher(self, resolved_id: int, domain: str, count: int) -> List[
...
+            else:
+                duplicates.append(r)
+        return uniques + duplicates
 
     @staticmethod
-    def _sample(recs, count):
-        """ Randomly sample articles or apply thompson sampling """
-        recs = random.sample(recs, count) if count < len(recs) else recs
-        return [CorpusRecommendationModel(corpus_item=r) for r in recs]
+    def _to_corpus_items(recs, count):
```
mmiermans, line 105: Nice rename. Much clearer. :+1:

app/data_providers/dispatch.py
```diff
@@ -78,13 +85,26 @@ async def by_publisher(self, resolved_id: int, domain: str, count: int) -> List[
...
+        seen_domains = set()
+        duplicates = []
+        uniques = []
+        for r in recs:
+            if r.domain not in seen_domains:
+                uniques.append(r)
+                seen_domains.add(r.domain)
+            else:
+                duplicates.append(r)
+        return uniques + duplicates
```
mmiermans, line 102: Similar logic is already implemented in `spread_publishers` and `spread_topics`. Might be good to standardize one a single approach to spread based on a generic attribute.

For this PR, can you move this function to `app/rankers/algorithms.py` so it's easy to keep track of the rankers that we have?

mmiermans (approved):
I only have some minor suggestions.

If you have a Jira ticket tracking this work, could you link it in the PR description and title?

app/data_providers/dispatch.py
eu9ene: I agree, I thought about this, but the requirements, use cases and input/output data structures are quite different with the home pub spread. I can move it to algorithms, yes

PR #1004, opened 2023-02-01
Filter original article from right rail recs

eu9ene:
# Goal

Fix a bug when an original article is shown on the right rail recommendations that look like a duplication of the main syndicated article.

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/TDP-307)


mmiermans (approved):
I think this would address the problem. :+1:

mmiermans:
In the past we solved this by filtering out articles that were very similar to the queried one, which feels like a more general solution. No need to pursue that as a solution for now.

eu9ene:
> In the past we solved this by filtering out articles that were very similar to the queried one, which feels like a more general solution. No need to pursue that as a solution for now.

Hm, yes it would be an interesting solution that also addresses duplicates with different IDs. We can potentially return the score from Qdrant and use it for this approach.

PR #1005, opened 2023-02-03
Add graphql caching for related

eu9ene:
# Goal

Move related recs caching to Apollo. Currently, not critical, but this will help to reduce the load on recAPI and vector store when firefox on-save and Android reader migration will be released.

I'm assuming here it will be cached on the Apollo server side. I don't completely understand how the whole thing works, so I might be wrong. I'll disable in-memory caching later if it works.

[Apollo docs](https://www.apollographql.com/docs/apollo-server/performance/caching/#using-with-federation)

[jira ticket](https://getpocket.atlassian.net/browse/TDP-304)

### DIFF
--- a/app/graphql/directives.py
+++ b/app/graphql/directives.py
@@ -1,0 +1,25 @@
+from enum import Enum
+
+import strawberry
+from strawberry import schema_directive
+from strawberry.schema_directive import Location
+
+
+@strawberry.enum()
+class CacheControlScope(Enum):
+    PUBLIC = strawberry.enum_value('PUBLIC')
+    PRIVATE = strawberry.enum_value('PRIVATE')
+
+
+@schema_directive(
+    locations=[
+        Location.FIELD_DEFINITION,
+        Location.INTERFACE,
+        Location.OBJECT
+    ],
+    name="cacheControl",
+    print_definition=True,
+)
+class CacheControl:
+    maxAge: int
+    scope: CacheControlScope = CacheControlScope.PUBLIC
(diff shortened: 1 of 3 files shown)


mmiermans (approved):
:+1: lgtm. Also a good opportunity for me to read up on how we've configured caching in [our gateway](https://getpocket.atlassian.net/wiki/spaces/PE/pages/1543995452/Pocket+Graph+Gateway).

eu9ene:
@bassrock will this work?

eu9ene:
> @eu9ene it depends on the query the client is making. This will do CDN level caching based on the lowest maxCacheAge when the client query is combined together at the gateway level. If the request the client makes calculates to `3600` max cache age, then it will be cached for `3600`. If they request one field that has a maxCacheAge of `5` it will only be cached at the cdn for `5` sec. This is documented in https://www.apollographql.com/docs/apollo-server/performance/caching/#using-with-federation

here is an example of a request:
```
 query GetSavedItemById($itemId: ID!) {
    user {
      savedItemById(id: $itemId) {
        ...SavedItemDetails
        annotations {
          highlights {
            id
            quote
            patch
            version
            _createdAt
            _updatedAt
            note {
              text
              _createdAt
              _updatedAt
            }
          }
        }
        item {
          ...ItemDetails
          ... on Item {
            article
            relatedAfterArticle(count: 3) {
              corpusRecommendationId: id
              corpusItem {
                thumbnail: imageUrl
                publisher
                title
                saveUrl: url
                id
                excerpt
              }
            }
          }
        }
      }
    }
  }
```

CorpusItem from curated-corpus subgraph doesn't have any cacheConrtol directives. Does it mean the whole request won't be cached? Do we have plans to make it cacheable?

I guess we can just deploy it and see.

PR #1007, opened 2023-02-10
Return empty recs for unsupported language

eu9ene:
# Goal

Fix a bug when related recommendations are displayed for a non-English item

## Jira

[ticket](https://getpocket.atlassian.net/browse/TDP-327)

## Implementation Decisions

@mmiermans @bassrock Will apollo provide `language` field to recAPI as a part of federated key if it is not present in the client request? I assume yes.
[edit]: it seems apollo is unhappy with this. any ideas on how to work around and pull language?

@anthony-liddle FYI

### DIFF
--- a/app/data_providers/dispatch.py
+++ b/app/data_providers/dispatch.py
@@ -7,7 +7,7 @@ from aws_xray_sdk.core import xray_recorder
 
 from app.config import DEFAULT_TOPICS, GERMAN_HOME_TOPICS
 from app.data_providers.corpus.corpus_feature_group_client import CorpusFeatureGroupClient
-from app.data_providers.item2item import Item2ItemRecommender, Item2ItemError, QdrantError
+from app.data_providers.item2item import Item2ItemRecommender, Item2ItemError, QdrantError, UnsupportedLanguage
 from app.data_providers.slate_providers.collection_slate_provider import CollectionSlateProvider
 from app.data_providers.slate_providers.for_you_slate_provider import ForYouSlateProvider
 from app.data_providers.slate_providers.life_hacks_slate_provider import LifeHacksSlateProvider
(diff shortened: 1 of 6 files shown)


bassrock:
@eu9ene For your use case I think you need to use computedFields to resolve it https://www.apollographql.com/docs/federation/entities-advanced#contributing-computed-entity-fields

So in your case _I think_ you would make relatedAfterArticle require `lang` (or whatever the field is on the Item object). then your resolver would get that field passed to it.

Summoning @kschelonka our resident resolver expert.

eu9ene:
> @eu9ene For your use case I think you need to use computedFields to resolve it https://www.apollographql.com/docs/federation/entities-advanced#contributing-computed-entity-fields
> 
> So in your case _I think_ you would make relatedAfterArticle require `lang` (or whatever the field is on the Item object). then your resolver would get that field passed to it.
> 
> Summoning @kschelonka our resident resolver expert.

I'm curious what's the difference between `Item` and `SyndicatedArticle`. For `SyndicatedArticle` I just added `publisherUrl originalItemId` as a part of key and it worked. `language` filed from `Item` in `parser` subgraph looks similar to `publisherUrl` field in the `SyndicatedArticle` in `syndication` subgraph. The only difference is that it's optional.

bassrock:
@eu9ene that resolvable piece is key, that is saying that Item is resolvable from your subgraph so federation does not pass those values through. You need it to not be resolvable so that Federation knows you require the data.

eu9ene:
> @eu9ene that resolvable piece is key, that is saying that Item is resolvable from your subgraph so federation does not pass those values through. You need it to not be resolvable so that Federation knows you require the data.

I tried this in my last commit and it now generates:

```
type Item @key(fields: "itemId language", resolvable: false) {
  itemId: String!
  language: String
```

However, apollo still complains with:

```
INVALID_FIELD_SHARING
Non-shareable field "Item.language" is resolved from multiple subgraphs: it is resolved from subgraphs "parser" and "recommendation-api" and defined as non-shareable in subgraph "parser"
```

bassrock:
@eu9ene yea this makes sense, tbh im surprised your publisherURL on SyndicatedArticle works, i think you need to do a computed field instead https://www.apollographql.com/docs/federation/entities-advanced#contributing-computed-entity-fields

bassrock:
Happy to pair live on this next week

eu9ene:
> @eu9ene yea this makes sense, tbh im surprised your publisherURL on SyndicatedArticle works, i think you need to do a computed field instead https://www.apollographql.com/docs/federation/entities-advanced#contributing-computed-entity-fields

I managed to generate a similar schema and validation passes for Item, but doesn't pass for SyndicatedArticle for some reason
```
SATISFIABILITY_ERROR

The following supergraph API query: { userById(id: "<any id>") { searchSavedItems( term: "A string value" filter: {} sort: {sortBy: CREATED_AT} pagination: {} ) { edges { node { savedItem { suggestedTags { savedItems( filter: {} sort: {sortBy: CREATED_AT, sortOrder: ASC} pagination: {} ) { edges { node { item { ... on Item { syndicatedArticle { relatedRightRail(count: 0) { ... } } } } } } } } } } } } } } cannot be satisfied by the subgraphs because: - from subgraph "syndication": cannot find field "SyndicatedArticle.relatedRightRail". - from subgraph "curated-corpus": cannot find field "SyndicatedArticle.relatedRightRail". - from subgraph "recommendation-api": cannot satisfy @require conditions on field "SyndicatedArticle.relatedRightRail".
```

I guess I'll just roll it back to use keys for now

mmiermans (approved):
:+1: lgtm

PR #1009, opened 2023-02-16
feat(Telemetry): Switch to OpenTelemetry [DIS-83]

mmiermans:
# Goal
Send telemetry to X-Ray and stop cluttering the logs with errors that X-Ray isn't working.

## Example trace
Steps to reproduce the below telemetry trace:
1. Enable the Pocket-Dev VPN
2. Run the 'RecommendationAPI / homeSlateLineup(en-US)' request in Postman, or the following request in an HTTP client of your choosing:
    ```shell
    curl --location --request POST 'https://recommendation-api.getpocket.dev' \
    --header 'userId: 123456' \
    --header 'encodedId: ABCDEF' \
    --header 'consumerKey: 123-456-678' \
    --header 'apiId: 123' \
    --header 'applicationName: Postman' \
    --header 'guid: 789' \
    --header 'Jwt: 1234567679065906570965' \
    --data-raw '{
        "query": "query PostmanHome { homeSlateLineup(locale: \"en-US\") { slates { headline recommendations(count: 3) { id corpusItem { id } } } }}",
        "variables": null,
        "operationName": "PostmanHome"
    }'
    ```
3. [Search for an X-Ray trace on port 8000](https://us-east-1.console.aws.amazon.com/xray/home?region=us-east-1#/traces?filter=http.url%20ENDSWITH%20%22%3A8000%2F%22)

[Example trace in Pocket-Dev](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#xray:traces/1-63efbec3-bb010469cbf74d51fbcd1091?~(query~(expression~'http.url*20ENDSWITH*20*22*3a8000*2f*22~filter~(node~'35d6c56d-ad47-56fc-8613-885d1795466c))~context~(timeRange~(delta~300000)~nodeVisualizationMode~'metrics))):

Screenshot from 2023-02-17 14-11-09

### First segment has request headers
The first segment is emitted by FastAPI and contains sanitized request headers:

Screenshot from 2023-02-17 10-30-42

### Second segment has GraphQL query
The second 'RecommendationAPI' segment contains the GraphQL query. I don't know why this isn't a child segment of the FastAPI one, but I think that doesn't matter much?

Screenshot from 2023-02-17 10-32-35

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/DIS-83

Docs
* [Choosing between the AWS Distro for OpenTelemetry and X-Ray SDKs](https://docs.aws.amazon.com/xray/latest/devguide/xray-instrumenting-your-app.html#xray-instrumenting-choosing)
  * Note: previously AWS 

## Implementation Decisions
- OpenTelemetry is [recommended by AWS](https://pypi.org/project/aws-xray-sdk/):
  > AWS X-Ray recommends using AWS Distro for OpenTelemetry (ADOT) to instrument your application instead of this X-Ray SDK due to its wider range of features and instrumentations.
- AWS, aiohttp, GraphQL and FastAPI are auto-instrumented. This allows us to get good insight into the application, without having to remember to put decorators everywhere.
- OpenTelemetry did not yet have instrumentation for aiobotocore, so I wrote custom instrumentation based on the [botocore instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore). The next step would be to open a PR against [open-telemetry/opentelemetry-python-contrib](https://github.com/open-telemetry/opentelemetry-python-contrib) to have this code merged in. I didn't want to block this PR on that.
- OpenTelemetry currently does not support async emitting of telemetry, but works on a separate thread and batches telemetry, so it shouldn't block the event loop. A load test showed no measurable performance difference between our current (broken) asynchronous X-Ray instrumentation and OpenTelemetry. The batch processor is configured to emit telemetry to a sidecar once every 5 seconds or 2048 spans. Telemetry is efficiently encoded in binary format using gRCP.

## Load test
The `dev.yml` load test was used to obtain the results below. It has a 30 minute ramp-up followed by a sustained 15 minute phase at 15 RPS.

There was no significant difference in latency or stability between this branch and the main branch. It's possible that the broken X-Ray telemetry in the main branch has a similar impact on performance as OpenTelemetry in this branch. I did not do a load test with all telemetry disabled.

#### X-Ray response time histogram

Screenshot from 2023-02-16 18-55-51

#### Artillery report

Note that the response times below include a network delay from my local machine to AWS.
```
All virtual users finished
Summary report @ 10:59:41(-0800) 2023-02-16
  Scenarios launched:  28911
  Scenarios completed: 28911
  Requests completed:  28911
  Mean response/sec: 10.7
  Response time (msec):
    min: 409.7
    max: 60356.2
    median: 515.5
    p95: 651.1
    p99: 1581.4
  Scenario counts:
    homeSlateLineup: 28911 (100%)
  Codes:
    200: 28909
    504: 2
```

**Cloudwatch Dashboard**
Screenshot from 2023-02-16 19-51-17

#### Artillery report

Note that the response times below include a network delay from my local machine to AWS.
```
All virtual users finished
Summary report @ 19:47:23(-0800) 2023-02-16
  Scenarios launched:  28767
  Scenarios completed: 28767
  Requests completed:  28767
  Mean response/sec: 10.65
  Response time (msec):
    min: 431.8
    max: 33069.5
    median: 524.2
    p95: 630.2
    p99: 753.6
  Scenario counts:
    homeSlateLineup: 28767 (100%)
  Codes:
    200: 28767
```

**Cloudwatch Dashboard**

Screenshot from 2023-02-16 19-50-05


eu9ene:
I would make sure we have a way to disable telemetry quickly if we see too much performance overhead.

docker-compose-dev.yml
```diff
@@ -10,6 +10,8 @@ services:
     env_file:
       - .docker/local-dev.env
```
eu9ene, line 11: let's make sure docker-compose.yml also works

Pipfile
```diff
@@ -42,9 +40,15 @@ aiomcache = "==0.6.*"
 ujson = "*"
 httptools = "*"
 aio-snowplow-tracker = "*"
-strawberry-graphql = {extras = ["fastapi"], version = "*"}
+strawberry-graphql = {extras = ["fastapi", "opentelemetry"], version = "*"}
 qdrant-client = "==0.11.6"
 pytz = "*"
+opentelemetry-api = "*"
```
eu9ene, line 46: it probably makes sense to pin those

app/instrumentation/aiobotocore/__init__.py
```diff
@@ -0,0 +1,297 @@
+# Based on botocore instrumentation:
```
eu9ene, line 1: Can we try to patch it instead of copy-pasting code? Contributing upstream might take some time.

eu9ene (approved):
We should make sure this thing doesn't break anything and doesn't bring too much overhead. There should be a way to disable it if we have issues. I think there should also be a plan to move the instrumentation code for aio libraries upstream in near future, otherwise, it will be hard to update them. 

Overall it looks good. We'll be able to understand what's happening when the request comes in and what parts to optimize. Especially if we plan to use recAPI for firefox new tab.

Great job, @mmiermans!

bkochendorfer:
This is awesome

PR #1014, opened 2023-03-14
feat(NewTab): Add newTabSlate [DIS-451]

mmiermans:
# Goal
Unblock development by Pocket's Firefox developers by adding a `newTabSlate` query that returns mock data.

This query won't be available to nightly users until we have localized content. Currently only English or German content is returned.

## Reference

Documentation:
* [newTabSlate API proposal](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2915368961)
* Shiproom (milestones, team, etc.)
* [technical spec](https://getpocket.atlassian.net/wiki/spaces/CP/pages/2924052481/Technical+Spec+Firefox+Recommendation+API)

Tickets:
* https://getpocket.atlassian.net/browse/DIS-451


app/data_providers/slate_providers/new_tab_slate_provider.py
```diff
@@ -0,0 +1,56 @@
...
+
+# Maximum tileId that Firefox can support. Firefox uses Javascript to store this value. The max value of a Javascript
+# number can be found using `Number.MAX_SAFE_INTEGER`. which is 2^53 - 1 because it uses a 64-bit IEEE 754 float.
+MAX_TILE_ID = 1 << 53 - 1
+# Generate tile_ids well out of the range of the MySQL-based system, which has a max tile_id of 99,999 as of 2023-03-13.
+# This is done to make it easy for engineers/analysts to see which system generated the identifier.
+MIN_TILE_ID = 100 * 100000
+
+
+class NewTabSlateProvider(SlateProvider):
```
eu9ene, line 18: I'm curious about terminology here. Why are we calling newtab a slate? I thought slate is part of a recommendation surface, like topic or Popular. Are there plans to support multiple slates on new tab? Or is is just for compatibility with our current home abstractions?

app/graphql/corpus_recommendation.py
```diff
@@ -8,5 +8,7 @@
 @strawberry.experimental.pydantic.type(model=CorpusRecommendationModel)
 class CorpusRecommendation:
     id: strawberry.ID
+    tile_id: strawberry.auto = strawberry.field(
```
eu9ene, line 11: it looks like tile_id is something very specific to new tab, why are we trying to reuse the Home domain model here instead of creating a new one?

app/models/corpus_item_model.py
```diff
@@ -4,3 +4,4 @@
 class CorpusItemModel(BaseModel):
     id: str
     topic: str = None
+    scheduled_date: str = None  # TODO: Query this value from scheduledSurface.items [DIS-452]
```
eu9ene, line 7: Same here, this is specific to new tab, we don't schedule anything for home or similar content recs, so this field doesn't make any sense for them.  I assume those different recommendation areas will grow in different directions even more over time. Why don't we create different minimal classes to support new tab.

app/data_providers/slate_providers/new_tab_slate_provider.py
mmiermans: A slate is a ranked list of related recommendations. The [Pocket glossary](https://docs.google.com/document/d/17d-NKQCEbSqkY-mnN1PSwwf4xsHc5nl67WbjKRR88QY/edit#) adds:
> These items should be related such that the slate makes sense. Examples include: Trending items, Long reads, Business (topic).

A slate lineup is "an ordered list of Slates to be shown to users in a specific context." There are no plans to support multiple slates on New Tab. To avoid confusion, we don't return a `CorpusSlateLineup` to clients when they should only expect a single slate.

app/graphql/corpus_recommendation.py
mmiermans: `CorpusRecommendation` is not necessarily tied to Home, but can be used for any recommendation made from the Corpus.

That said, you have a great point that Firefox is the only client that would need a `tile_id`, and it would be cleaner to have a separate type for this. :writing_hand: Let me take a stab at that..

app/models/corpus_item_model.py
mmiermans: Good catch. `scheduled_date` is a property of [ScheduledSurfaceItem](https://studio.apollographql.com/graph/pocket-client-api/schema/reference/scalars/Date?query=scheduled&variant=current), not CorpusItem.

app/graphql/corpus_recommendation.py
mmiermans: After going down the path of separating types and reflecting on this, I lean towards keeping `tile_id` in `CorpusRecommendation`:
- Adding a `NewTabCorpusRecommendation` type also requires a separate `NewTabCorpusSlate` type, a separate `NewTabCorpusRecommendation` resolver, etc. Additional changes might also be required in Snowplow JSON schema and Dbt modeling.
- [DIS-491](https://getpocket.atlassian.net/browse/DIS-491) `tileId` is deprecated and I am going to work with Firefox Devs and Firefox Data engineering to replace it with `CorpusRecommendation.id`. This will bring Firefox New Tab in line with how telemetry is identified on Home.
- One of the advantages of GraphQL is that clients can have great freedom in querying exactly the information they need. My first thought was that Firefox would not benefit from any extensions to `CorpsuRecommendations` because its product requirements are set in stone. But maybe it's also the other way around, in that New Tab product requirements have not changed much because the existing Rest API was so hard to change?

@eu9ene  Are you ok with keeping the existing strategy of using a deprecation message to prevent any clients except Firefox New Tab from using `tileId`, and removing this field over time?

eu9ene: Yeah, unification and having the same pipelines for all recs seems like a good strategy too. Let's add the depreciation flag and maybe some other attributes or comments to let the clients know that some fields are relevant only for specific scenarios. My concern here is that when deployed, this API will be hard to change, so we should plan for the future.

mmiermans: :+1: I expanded the deprecation reason on the line below to this:

> Only to be used by Firefox. Other clients should use `id`. We plan to also migrate Firefox New Tab to use CorpusRecommendation.id instead of tileId to track recommendation telemetry.

I appreciate your focus on the API. In hindsight I should've included you for review on the [API proposal](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2915368961/Pocket+New+Tab+Recommendations+for+New+Markets). I'll try to make sure to remember to include you in the future.

eu9ene: That's why it might be a good idea to do those proposals as a part of a pull request. They might look fine in confluence but be inconsistent with the rest of the code.

eu9ene (approved):

PR #1026, opened 2023-04-19
chore(Synthetic): rename to clarify it targets New Tab [DIS-579]

mmiermans:
# Goal
Rename the synthetic and the alarm to clearly indicate that they target the New Tab endpoint in Recommendation API.

## Todos
- [x] Test deployment in Pocket-Dev

## Reference

Slack thread:
* https://pocket.slack.com/archives/C02KH5U7G79/p1681843679320649

Jira ticket:
* https://getpocket.atlassian.net/browse/DIS-579


.aws/src/files/python/new_tab_synthetic.py
```diff
@@ -67,7 +67,7 @@ def main():
         "variables": {
             "locale": "en-US",
             "region": "US",
-            "count": 50,
+            "count": 30,
```
mmiermans, line 70: Matches the number that Firefox requests

bkochendorfer (approved):

PR #1028, opened 2023-04-24
Use only recent articles for non-syndicated similar recs

eu9ene:
# Goal

Fix recommending outdated news in Reader mode and for Firefox on-save.

It turned out we cannot rely on the time-sensitive flag to identify evergreen content, so I have to add a hard filter.

## Reference

Tickets:
* [jira](https://getpocket.atlassian.net/browse/TDP-351)

## Implementation Decisions
- use 90 days as a time window that seems reasonable not to recommend too outdated content but still preserve some content pool (~2.5K items)
- keep recommendations on the syndicated page as is because of obligations with the publishers, also the syndicated articles are supposed to be evergreen. Some publisher-based articles might be old though
- we use first_parsed_date as a timestamp in Qdrant, which seems to be the most reliable way to estimate the article's publishing date, it can be way off of course.

### DIFF
--- a/app/data_providers/item2item.py
+++ b/app/data_providers/item2item.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime, timedelta
 from typing import List, NamedTuple
 
 from aiocache import cached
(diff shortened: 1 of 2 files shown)


mmiermans (approved):
👍lgtm

PR #1030, opened 2023-05-02
feat(NewTab): Spread publishers and fix rollover time [DIS-581][DIS-593]

mmiermans:
# Goal
- [DIS-581](https://getpocket.atlassian.net/browse/DIS-581) spread recs from the same publisher 6 apart, per the [New Tab slate spec](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2927100008/Fx+NewTab+Slate+spec+for+New+Markets).
- [DIS-593](https://getpocket.atlassian.net/browse/DIS-593) content for new markets should roll over at midnight CET.

## Implementation Decisions
- Consolidate `spread_topics` and `spread_publishers`. I did not consolidate `unique_domains_first` because I'm not totally sure what the expected behavior is, and it's not covered by tests. Probably we can replace it with `unique_domains_first = partial(spread_attribute, name='domain')`.
- Increase test coverage for the `NewTabSlateProvider.rank_corpus_items`.

[DIS-581]: https://getpocket.atlassian.net/browse/DIS-581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIS-593]: https://getpocket.atlassian.net/browse/DIS-593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/mocks/feature_store_mock.py
+++ b/tests/mocks/feature_store_mock.py
@@ -90,9 +90,9 @@ class FeatureStoreMock:
                     errors.append({'ErrorMessage': str(e)})
 
         if errors:
-            return {'Records': [], 'Errors': errors}
+            return {'Records': records, 'Errors': errors}
         else:
-            return {'Records': []}
+            return {'Records': records}
 
     def _put_record_stub(self, *args, **kwargs):
         feature_group_name = kwargs['FeatureGroupName']
--- a/tests/unit/data_providers/test_new_tab_slate_provider.py
+++ b/tests/unit/data_providers/test_new_tab_slate_provider.py
@@ -70,7 +70,41 @@ class TestNewTabSlateProvider:
         assert not any(r.levelname == 'ERROR' for r in caplog.records)
 
     @pytest.mark.parametrize('repeat', range(10))  # Thompson sampling is non-deterministic, so repeat the test.
-    async def test_publisher_spread(
+    async def test_rank_by_scheduled_date_rank(
+            self, new_tab_slate_provider, corpus_items_10, aiocache_functions_fixture, repeat):
+        ranked_items = await new_tab_slate_provider.rank_corpus_items(items=corpus_items_10)
+
+        assert len(corpus_items_10) == len(ranked_items)
+        # Assert items are ranked by scheduled date in descending order
+        last_scheduled_date = None
+        for corpus_item in ranked_items:
+            scheduled_date = new_tab_slate_provider.corpus_api_client.get_scheduled_date(corpus_item.id)
+            if last_scheduled_date is not None:
+                assert last_scheduled_date >= scheduled_date
+            last_scheduled_date = scheduled_date
+
+    async def test_rank_by_thompson_sampling(
+            self, new_tab_slate_provider_without_scheduled_date, corpus_items_10, aiocache_functions_fixture):
+        # Make all publishers unique, to ensure publisher spreading has no effect on the ranking.
+        for i, corpus_item in enumerate(corpus_items_10):
+            corpus_item.publisher = f'Publisher {i}'
+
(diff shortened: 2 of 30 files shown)


tests/mocks/feature_store_mock.py
```diff
@@ -90,9 +90,9 @@ def _batch_get_record_stub(self, *args, **kwargs):
                     errors.append({'ErrorMessage': str(e)})
 
         if errors:
-            return {'Records': [], 'Errors': errors}
+            return {'Records': records, 'Errors': errors}
         else:
-            return {'Records': []}
+            return {'Records': records}
```
mmiermans, line 95: This was a bug in the mock feature store that went unnoticed because it was only used with `empty_feature_group.json`.

tests/unit/data_providers/test_new_tab_slate_provider.py
```diff
@@ -61,6 +68,55 @@ async def test_rank_corpus_items(self, new_tab_slate_provider, corpus_items_10,
...
+            ranked_items = await new_tab_slate_provider_without_scheduled_date.rank_corpus_items(items=corpus_items_10)
+            top_ranked_ids.append(ranked_items[0].id)
+            bottom_ranked_ids.append(ranked_items[-1].id)
+
+        # The NewTabSlateProvider fixture receives engagement data from corpus_engagement.json, which has a 5% CTR for
+        # id=7 with 10,000 impressions. The prior is lower than 5%, so id 7 should be ranked first most often.
+        most_frequent_top_ranked_id = max(set(top_ranked_ids), key=top_ranked_ids.count)
+        assert most_frequent_top_ranked_id == '7'
+        # All other items are ranked by the prior, so the bottom ranked item should vary.
+        assert len(set(bottom_ranked_ids)) > 3  # Should be close to 9 with high probability.
```
mmiermans, line 103: I ran this test 1,000 times to confirm these assertions pass reliably, but I didn't do the math. I believe the 5% CTR in the fixture data should ensure it passes reliably, even when the priors are adjusted.

tests/unit/rankers/test_algorithms.py
```diff
@@ -145,11 +145,11 @@ def test_spread_publishers_cannot_spread(self):
         recs[2].publisher = 'donnie.com'
         recs[3].publisher = 'donnie.com'
         recs[4].publisher = 'thedude.com'
-        recs[5].publisher = 'thedude.com'
-        recs[6].publisher = 'abides.com'
+        recs[5].publisher = 'abides.com'
+        recs[6].publisher = 'thedude.com'
```
mmiermans, line 149: This seems like a bug in the old `spread_publishers` code. Moving `thedude.com` down one position increases the distance to the rec with the same domain above it, without bringing `abides.com` within distance 3.

eu9ene (approved):
Looks good but I suggest including high-level functional tests to check that the main requirements for the new tab slate are satisfied.

mmiermans:
Capturing here what we discussed in Slack for future reference. Synthetic monitors are probably a better tool to achieve confidence that the requirements are met end-to-end. I like the idea of adding a synthetic API monitor that checks publisher diversity and recency.

I think functional API tests are not the right tool to test publisher spreading or Thompson sampling, because it's hard to observe those from an API response. Recommendation API does not return publishers, and Thompson sampling introduces a random variable. Unit tests can eliminate the randomness by running tests 100 times in a fraction of a second, but functional API tests take about a second to execute.

PR #1031, opened 2023-05-08
fix(ECS): Increase minimum task count

mmiermans:
# Goal
Avoid spikes in latency on Recommendation API such as happened on May 8th 5:33 am.

## Reference

- [Jira ticket](https://getpocket.atlassian.net/browse/DIS-675)
- Alert on Slack

## Implementation Decisions
- I don't immediately see the root cause. This was something quick I thought we could try. Further investigation is needed if this starts happening more frequently.

## Evidence
Latency spike at 5:30 AM PST corresponded with an increase in requests.
Screenshot from 2023-05-08 12-41-13

### Home on Web
Request increase seems to correspond with latency increase.

Screenshot from 2023-05-08 12-54-18

### Topic explore pages
Screenshot from 2023-05-08 12-56-40

### Firefox New Tab for new markets
Screenshot from 2023-05-08 12-56-32

### Web Discover
Screenshot from 2023-05-08 12-56-24

### Android Home
Screenshot from 2023-05-08 12-56-03

### DIFF
--- a/.aws/src/main.ts
+++ b/.aws/src/main.ts
@@ -336,7 +336,7 @@ class RecommendationAPI extends TerraformStack {
             },
 
             autoscalingConfig: {
-                targetMinCapacity: 2,
+                targetMinCapacity: 4,
                 targetMaxCapacity: 10
             },
             alarms: {


mmiermans:
The next time this happens it might be useful to look at [Recommendation API X-Ray traces](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#xray:traces/query?~(query~(filter~()~expression~'service*28id*28name*3a*20*22RecommendationAPI*22*2c*20type*3a*20*22AWS*3a*3aECS*3a*3aContainer*22*29*29)~context~())). They're only kept around for 6 hours, and it's been longer than that since latency spiked.

revelyn-mozilla (approved):
bump asg

bkochendorfer:
> Request increase seems to correspond with latency increase.
That's interesting, seems like we might need an exponential backoff in the client?

bkochendorfer (approved):

mmiermans:
> > Request increase seems to correspond with latency increase.
>
> That's interesting, seems like we might need an exponential backoff in the client?

Maybe? Or maybe there's a performance bug in Recommendation API?

PR #1033, opened 2023-05-09
feat: emit ranked_with_engagement_updated_at and refactor tests

mmiermans:
# Goal
Emit `ranked_with_engagement_updated_at` for recommendations ranked with Thompson sampling, such that we can report and alert on the delay in our engagement pipeline. This timestamp should be set to the `UPDATED_AT` value fetched from the engagement Feature Group.

## Reference

* https://getpocket.atlassian.net/browse/DIS-587
* [NewTab Reliability Doc](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2927099995/Fx+NewTab+for+New+Markets+Reliability+Doc)

## Implementation Decisions
- Point Snowplow tests to Snowplow Micro instead of a custom mocked server. This makes tests simpler and improves coverage.
- Add a healthcheck to docker-compose.yml to allow tests to pass in CircleCI. `test_snowplow_corpus_recommendations_tracker.py` happened to be the first tests to run, and they [were failing](https://app.circleci.com/pipelines/github/Pocket/recommendation-api/4048/workflows/cf55b9ce-21a4-43de-9674-c6a6a3ce95d2/jobs/19439/tests) because Snowplow Micro was not yet available at that time.
- Set `ranked_with_engagement_updated_at` in the `thompson_sampling` ranker to ensure it's present on all recommendations (Home and NewTab) where Thompson sampling is used.


docker-compose.yml
```diff
@@ -19,10 +19,14 @@ services:
...
-      - snowplow
-      - qdrant
+      localstack:
+        condition: service_started
+      memcached:
+        condition: service_started
+      snowplow:
+        condition: service_healthy  # snowplow needs to be healthy before tests can start.
+      qdrant:
+        condition: service_started
```
mmiermans, line 29: `service_started` is the default. [Docker Compose depends_on documentation](https://docs.docker.com/compose/compose-file/05-services/#depends_on) states that this is the default in the short syntax that was used before.

app/models/corpus_item_model.py
```diff
@@ -1,7 +1,17 @@
...
+
+from pydantic import BaseModel, Field
 
 
 class CorpusItemModel(BaseModel):
     id: str
     topic: str = None
     publisher: str = None
+
+    ranked_with_engagement_updated_at: Optional[datetime] = Field(
```
eu9ene, line 12: I'm a bit out of context but recAPI doesn't look like the right place to monitor issues in our data pipelines. Approving anyway

eu9ene (approved):

app/models/corpus_item_model.py
mmiermans: :thinking: Maybe I should've considered asking Braun for review, because he and I discussed how we could implement our SLIs. I'll think about that next time.

mmiermans: We want the recommendation freshness SLI to cover the entire pipeline responsible for Thompson sampling. recAPI is the last link in the chain, so we need to emit some data here to know which engagement data was used.

PR #1034, opened 2023-05-09
fix: make publisher spreading the primary sort order

mmiermans:
# Goal
Make publisher spreading the primary sort order to spread recommendations from the same publisher should be spread 6 positions apart.

Before, publisher spreading did not work correctly because items of the same publisher could be far apart before sorting on recency, and close together afterwards.

## Review
- [ ] @lfishhead Could you review the [NewTab slate spec](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2927100008/Fx+NewTab+Slate+spec+for+New+Markets?atlOrigin=eyJpIjoiZDE4OWE0MzVjYTU4NDE1YWE0MTJjZjUwZDEyZmRlMDgiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ) and check the sign-off box at the bottom of the doc?

## Deployment
- [x] Wait until Sam and Carolyn have signed off on the [Slate spec](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2927100008/Fx+NewTab+Slate+spec+for+New+Markets?atlOrigin=eyJpIjoiZDE4OWE0MzVjYTU4NDE1YWE0MTJjZjUwZDEyZmRlMDgiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ).

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/DIS-681

Slack thread:
* https://pocket.slack.com/archives/C04LKH4J1M0/p1683672233845539

## Implementation Decisions
- This did not get caught in tests because I had isolated the effects of the rankers in tests. I changed `test_rank_by_publisher_spread` to include scheduled dates to test the combined effect of sorting on recency and pub spread. I'm open to suggestions on how to test better?


app/data_providers/slate_providers/new_tab_slate_provider.py
```diff
@@ -104,10 +104,11 @@ async def rank_corpus_items(self, items: List[CorpusItemModel], *args, **kwargs)
             default_alpha_prior=188,  # beta * P99 German NewTab CTR for 2023-03-28 to 2023-04-05 (1.5%)
             default_beta_prior=12500)  # 0.5% of median German NewTab item impressions for 2023-03-28 to 2023-04-05.
 
-        # 2. Secondary sort order is publisher diversity.
-        items = spread_publishers(items, spread_distance=PUBLISHER_SPREAD_DISTANCE)
-
-        # 1. Primary sort order is recency. Sort is stable, so it will preserve existing order within a scheduled date.
+        # 2. Secondary sort order is recency.
```
lfishhead, line 107: I would be concerned this may negatively impact CTR.  My understanding is that there is a time decay applied somewhere in the feed machine such that most items drop after a day.  Depending on how many items you are loading here, I might suggest recency third and TS second.  Additionally, I recall there is some impression capping on the firefox side, so I think the recency concerns may be more limited.

lfishhead (approved):
no code changes, but I'd encourage a conversation about this with product owners and am happy to be part of that

app/data_providers/slate_providers/new_tab_slate_provider.py
mmiermans: This was exactly the feedback I was looking for. TS is not a stable sort, because all values it sorts on are distinct, so any sort that would happen before it would not have any effect.

A penalty approach would be totally in our power to implement. Let's have the conversation!

mmiermans: Captured this in [DIS-686](https://getpocket.atlassian.net/browse/DIS-686) so it won't get lost.

PR #1038, opened 2023-05-23
Switch to new CF model artifacts

eu9ene:
- Update s3 location to use the models that are updated daily
- Implement a sanity check using test predictions from Metaflow

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/DIS-747)


chidamnat (approved):
LGTM. Thanks!

PR #1040, opened 2023-05-25
Add more guardrails for hybrid CF model loading

eu9ene:
# Goal

- Make sure the RecAPI doesn't crash if the HybridCF model suddenly becomes too large for the current setup
- Make model loading optional to still serve the default recs if something went wrong with the model in s3. Other functionality of RecAPI should also work

@chidamnat Let's do a code freeze on Metaflow side for the `v1` model. If we want to improve it let's introduce `v2` and then we'll go through testing again before redeploying it. Even though we'll have some guardrails there shouldn't be any surprises on RecAPI side regarding the updated model. It should be exactly the same setup as was tested during the deployment.

Tickets:
* [Link to JIRA tickets](https://getpocket.atlassian.net/browse/DIS-755)

### DIFF
--- a/app/recommenders/hybrid_cf.py
+++ b/app/recommenders/hybrid_cf.py
@@ -47,6 +47,10 @@ class HybridCFRecommender:
     """
     Hybrid Collaborative Filtering recommender
     """
+    # 230 Mb in the storage utilizes ~30% of RecAPI memory (8GB) (!)
+    # after loading objects take ~650MB per worker x2 = 1300MB + some leakage
+    MAX_STORAGE_ARTIFACTS_SIZE = 500 * 1024 * 1024  # 500Mb
+
     def __init__(self, model_loader: ModelLoader):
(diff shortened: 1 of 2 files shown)


app/recommenders/hybrid_cf.py
```diff
@@ -47,6 +47,10 @@ class HybridCFRecommender:
     """
     Hybrid Collaborative Filtering recommender
     """
+    # 230 Mb in the storage utilizes ~30% of RecAPI memory (8GB) (!)
+    # after loading objects take ~650MB per worker x2 = 1300MB + some leakage
+    MAX_STORAGE_ARTIFACTS_SIZE = 500 * 1024 * 1024  # 500Mb
```
chidamnat, line 52: as discussed with @eu9ene , my recommendation is to bump this up to 600 / 700 MB on the safer side. Thanks!

PR #1039, opened 2023-05-24
Add support of hasUserModel Unleash strategy

eu9ene:
# Goal

Being able to enroll in an experiment only users that are supported by the model

- [x] tested on dev feature flags + recAPI

Tickets:
*[ Link to JIRA tickets](https://getpocket.atlassian.net/browse/DIS-756)

## Implementation Decisions

Use the existing `hasUserModel` strategy that was developed for recit. It works well here, the only downside that it says recit in property name.

PR #1041, opened 2023-05-25
Add support of hasUserModel Unleash strategy

eu9ene:
# Goal

Being able to enroll in an experiment only users that are supported by the model

- [x] tested on dev feature flags + recAPI

Tickets:
*[ Link to JIRA tickets](https://getpocket.atlassian.net/browse/DIS-756)

## Implementation Decisions

Use the existing `hasUserModel` strategy that was developed for recit. It works well here, the only downside that it says recit in property name.


eu9ene:
it's the same as https://github.com/Pocket/recommendation-api/pull/1039 that was created from the the dev branch

mmiermans (approved):

PR #1044, opened 2023-06-02
feat: Sentry transaction is set to graphql operationName

mmiermans:
# Goal
Add the GraphQL operation name as a searchable tag to Sentry. This has several benefits:
1. It's immediately clear which operation causes a Sentry error, because Sentry will show a histogram of operations with each issue.
2. We can search for operations in Sentry.
3. We can define different alert rules or thresholds for particular operations.

## Before
`transaction` was always set to `/`, because all GraphQL queries are served from this route. There was no way to search by the GraphQL query or operation in Sentry. The 'body' field does contain the operation name, but this field is not indexed.

image

## After
`transaction` is set to the GraphQL operation name. This is a [searchable property](https://docs.sentry.io/product/sentry-basics/search/searchable-properties/) in Sentry, so we can now [search](https://pocket.sentry.io/issues/?limit=5&project=5503693&query=is%3Aunresolved+transaction%3ANewTabSlate&referrer=performance-related-issues&sort=new&statsPeriod=14d) and [alert](https://pocket.sentry.io/alerts/rules/recommendation-api/14455029/) for specific GraphQL operations.

image

The Sentry transaction is a tag that alerts can filter on:

Screenshot from 2023-06-02 14-54-22

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/DIS-791

PR against sentry-python:
* https://github.com/getsentry/sentry-python/pull/2150

## Implementation Decisions
- https://github.com/getsentry/sentry-python/pull/2150 adds a searchable tag to the Python sentry-sdk library. If this is merged and released, we can update sentry-sdk and remove this custom Sentry integration.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -13,12 +13,12 @@ from opentelemetry.sdk.resources import Resource, get_aggregated_resources
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 from opentelemetry.sdk.trace.sampling import TraceIdRatioBased
-from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from strawberry.fastapi import GraphQLRouter
 
 from app.cache import initialize_caches
 from app.config import ENV, ENV_PROD, sentry as sentry_config, log_level, otel_daemon_address
 from app.graphql.graphql_router import schema
+from app.graphql.sentry_graphql_middleware import GraphqlIntegration
 from app.health_status import get_health_status, set_health_status, HealthStatus
 from app.instrumentation.aiobotocore import AiobotocoreInstrumentor
 from app.models.candidate_set import candidate_set_factory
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -32,13 +32,13 @@
...
+    environment=sentry_config['environment'],
+    integrations=[
+        GraphqlIntegration(),
+    ],
 )
-# Ignore graphql.execution.utils to prevent duplicate Sentry events. Exceptions are handled by GraphQLSentryMiddleware.
-sentry_sdk.integrations.logging.ignore_logger("graphql.execution.utils")
 
 app = FastAPI()
-app.add_middleware(SentryAsgiMiddleware)
```
mmiermans, line 41: This middleware was no longer necessary because Sentry integrates with FastAPI by default now.

app/main.py
```diff
@@ -32,13 +32,13 @@
...
     dsn=sentry_config['dsn'],
     release=sentry_config['release'],
-    environment=sentry_config['environment']
+    environment=sentry_config['environment'],
+    integrations=[
+        GraphqlIntegration(),
+    ],
 )
-# Ignore graphql.execution.utils to prevent duplicate Sentry events. Exceptions are handled by GraphQLSentryMiddleware.
-sentry_sdk.integrations.logging.ignore_logger("graphql.execution.utils")
```
mmiermans, line 38: This was added a long time ago when we used a different GraphQL server library. I tried removing it and don't see duplicate events.

mmiermans:
Sentry is planning GraphQL improvements that would make this change superfluous:
> We're currently looking into improving our GraphQL support in general (better errors and spans) so there are changes on the way that will supersede this PR, adding better support for GraphQL queries not just in Starlette but other frameworks as well.

PR #1047, opened 2023-06-09
drop the sessionId from the unleash query

chidamnat:
# Goal
Some users are assigned to more than one treatment variant.
https://pocket.slack.com/archives/C02JZ4TRF0S/p1686265277754779?thread_ts=1686251034.146639&cid=C02JZ4TRF0S

One of the suspects is the presence of sessionId in the unleash call. We wanted to verify if that is helpful

## Todos
- [x] Outstanding todo
- [x] Completed todo

## Review
- [x] Security @ReviewerA
- [x] Test coverage @ReviewerC
- [ ] ...

Have you followed the guidelines in our Contributing document?

## Deployment
- [ ] Secrets?

## Reference

Tickets:
* Link to JIRA tickets

### DIFF
--- a/app/data_providers/unleash_provider.py
+++ b/app/data_providers/unleash_provider.py
@@ -85,7 +85,6 @@ class UnleashProvider:
                     'appName': self.unleash_config.APP_NAME,
                     'environment': self.unleash_config.ENVIRONMENT,
                     'userId': user.hashed_user_id,
-                    'sessionId': user.hashed_guid,
                     'properties': {
                         'locale': user.locale
                     }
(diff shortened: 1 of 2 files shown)


mmiermans:
Slack thread for future reference: https://pocket.slack.com/archives/C02JZ4TRF0S/p1686265277754779?thread_ts=1686251034.146639&cid=C02JZ4TRF0S

> Try removing sessionId from the [context that we pass to Unleash](https://github.com/Pocket/recommendation-api/blob/main/app/data_providers/unleash_provider.py#L88). I wouldn't expect this to change anything, but it's worth a try to see if this solves the issue. sessionId is only useful for logged-out experiments, which we don't plan to do anyway.

PR #1049, opened 2023-06-15
Increase available memory

eu9ene:
# Goal

Fix ratio memory/core after deployment reconfiguration. We anticipate the ML model that takes most of the space to grow in the next version. Also, I think it's better to be on the safe side and have lower memory utilization than 70%. The model is reloaded periodically and updated by metaflow. It can suddenly become slightly larger. Even though we have some checks for its size in the reloading code we do not want to end up with OOM, it's better to leave some buffer.

@nathan-barrett was this reconfiguration done because of the cost or just to optimize resource utilization?

Screenshot 2023-06-15 at 10 29 02 AM

### DIFF
--- a/.aws/src/main.ts
+++ b/.aws/src/main.ts
@@ -159,7 +159,7 @@ class RecommendationAPI extends TerraformStack {
             domain: config.domain,
             taskSize: {
                 cpu: 4096,
-                memory: 8192,
+                memory: 16384,
             },
             containerConfigs: [
                 {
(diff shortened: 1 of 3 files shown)


eu9ene:
Something weird happened to CircleCI docker build step: it fails on `pip install` with `RuntimeError: can't start new thread`

n0coast (approved):

revelyn-mozilla (approved):
fix validated with successful run and awaiting signoff from devs team

jeshuaborges (approved):

PR #1051, opened 2023-06-28
Increase memory

eu9ene:
# Goal

Fix ratio memory/core after deployment reconfiguration. We anticipate the ML model that takes most of the space to grow in the next version. Also, I think it's better to be on the safe side and have lower memory utilization than 70%. The model is reloaded periodically and updated by metaflow. It can suddenly become slightly larger. Even though we have some checks for its size in the reloading code we do not want to end up with OOM, it's better to leave some buffer.

Also, update CI orb.

This is a copy of this PR https://github.com/Pocket/recommendation-api/pull/1049 which I couldn't land because of the unsigned commits.

### DIFF
--- a/.aws/src/main.ts
+++ b/.aws/src/main.ts
@@ -159,7 +159,7 @@ class RecommendationAPI extends TerraformStack {
             domain: config.domain,
             taskSize: {
                 cpu: 4096,
-                memory: 8192,
+                memory: 16384,
             },
             containerConfigs: [
                 {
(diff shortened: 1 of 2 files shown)


jeshuaborges (approved):

eu9ene:
Updated to 12GB instead of 16 since it's not clear whether we need to accommodate a larger HybridCF model in near future

n0coast (approved):

PR #1052, opened 2023-07-05
Add start period to the healthcheck

eu9ene:
# Goal

The deployment fails because it takes about a minute to start the app with the new model loading. 

Let's add a 60-second starting period before the health check kicks in.


mmiermans (approved):
:+1: A minute of startup time feels reasonable to me.

izo3 (approved):

PR #1055, opened 2023-07-07
feat: Enable automatic updates from renovate

mmiermans:
# Goal
Automatically update Python packages to:
1. Install security updates in a timely manner.
2. Keep libraries up to date, and not get stuck on old versions with many breaking changes.

## Deployment
- [ ] [renovate](https://github.com/apps/renovate) is allowed on your repository
- [ ] [renovate-approve](https://github.com/apps/renovate-approve) is allowed on your repository
- [x] All dependabot and auto-merge files are removed from your repository. While, not necessary, they will conflict with each other.
- [ ] Allow auto-merge is enabled in your GitHub repository settings. While not necessary this will allow Renovate to use Github's default auto-merging strategy.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/DIS-576

Pocket's Renovate config:
* https://github.com/Pocket/renovate-config


eu9ene:
I understand the idea but python package ecosystem is very brittle and usually I prefer not to touch the dependencies when things work. Updating one package can cause ugly conflicts and since they are open source there are no guarantees of backward compatibility. What security risks do you use for RecAPI taking into account that it's behind the VPN and accessed only by internal services?

mmiermans:
I think Evgeny makes some valid points, and for some repositories locking down dependencies would be the best strategy. Here is why I think it is better to update packages on recommendation-api more frequently:

1. For the last year or so, we've been relying on test coverage alone to add/update packages in this repo. I don't think we've had any incidents with breaking changes in package updates.
2. There is already a small risk today that our test coverage doesn't catch a breaking change when we add/update a package, because we don't pin all packages today. Automatic updates don't change that, but they do make it easier to find the cause. We will be notified when this happens by Sentry alerts and API monitoring for NewTab.
3. While a VPN limits the attack surface, it doesn't eliminate it. It's less effort to update packages than to investigate whether a security bug can be exploited.
4. Working through conflicts isn't avoidable in an actively used repo. The question is only whether we want to work through them gradually as part of continuous delivery, or only as needed. For recommendation-api I think the former is worth the investment.

mmiermans:
We decided that weekly updates with human review would offer the best balance between avoiding an update backlog, and minimizing risk from breaking changes.

However, I don't see this as a priority now, so am closing this PR for now.

PR #1058, opened 2023-08-02
fix: newTabSlate timezone depends on scheduled surface

mmiermans:
# Goal
New Tab should roll over based on the ScheduledSurface's timezone.

## Reference

Tickets:
* https://getpocket.atlassian.net/browse/DIS-853

List of ScheduledSurfaces and corresponding timezones:
* https://github.com/Pocket/curated-corpus-api/blob/main/src/shared/types.ts

## Implementation Decisions
- Add test coverage for `CorpusApiClient`.
- Add missing dependency `aiohttp` to Pipfile. It was already installed as a sub-dependency, but we depend directly on it. This is unrelated to PR goal, and should probably have gone in a separate PR, but I noticed it as I was going.
- I've hardcoded timezones here for now, because curated-corpus-api doesn't expose them yet.


jeshuaborges (approved):
Nice work. The tests look great.

PR #1060, opened 2023-08-08
chore: migrate to new user preferences store

mmiermans:
# Goal
Cleanup code by only reading/writing recommendation preferences based on hashed_user_id.

# Context
We’ve been double-writing recommendation (topic) preferences to two stores keyed on user_id and hashed_user_id since August 2022. The latter is the recommended identifier for users. Last year we also did a data migration from the old to the new store.

We also did a data migration in 2022 from the old to the new store. The Athena tables have an almost identical number of distinct user ids. The difference is small enough that I don't think it warrants further investigation.

|                | **unique user count** |
|:--------------:|:---------------------:|
|     **old**    | 557953                |
|     **new**    | 557568                |
| **difference** | 385 (0.07%)                   |

Athena query:
```sql
SELECT 'old', count(distinct user_id) FROM "sagemaker_featurestore"."production-user-recommendation-preferences-v1-1654798646"
UNION ALL
SELECT 'new', count(distinct hashed_user_id) FROM "sagemaker_featurestore"."production-user-recommendation-preferences-v2-1660003138";
```

## Deployment
- [ ] Merge and deploy change
- [ ] Delete old Feature Group

## Reference

Tickets:
* https://mozilla-hub.atlassian.net/browse/MC-26

## Implementation Decisions
- I kept 'V2' in the class name to make the history easier to read.


app/graphql/resolvers/corpus_slate_lineup_resolvers.py
```diff
@@ -66,7 +66,7 @@ async def resolve_home_slate_lineup(root, info: Info, locale: str = 'en-US') ->
 
         slate_lineup_model = await HomeDispatch(
             corpus_client=di.corpus_feature_group_client,
-            preferences_provider=UserRecommendationPreferencesProvider(di.aioboto3_session, topic_provider),
+            preferences_provider=UserRecommendationPreferencesProviderV2(di.aioboto3_session, topic_provider),
```
jeshuaborges, line 69: Should we rename `UserRecommendationPreferencesProviderV2` to `UserRecommendationPreferencesProvider` as future devs may not care, or know about the version history?

app/graphql/resolvers/user_recommendation_preferences_resolvers.py
```diff
@@ -42,21 +30,12 @@ async def update_user_recommendation_preferences(
...
         hashed_user_id=request_user.hashed_user_id,
         updated_at=datetime.datetime.utcnow(),
         preferred_topics=preferred_topics
     )
 
-    await asyncio.gather(
-        preferences_provider.put(model),
-        preferences_provider_v2.put(model_v2),
-    )
+    await preferences_provider_v2.put(model_v2)
```
jeshuaborges, line 39: Same comment about the _v2 stuff here.

jeshuaborges, line 39: Also nice work reducing to a single path.

jeshuaborges (approved):
Nice work. This all looks super straight forward to me. My only feedback would be to rename everything that ended up with a 'v2' suffix to remove the suffix.

jsrawan-mobo (changes requested):
When I searched for 'user.user_id' i found following files with references that I don't see in the PR

app/data_providers/snowplow/entities.py
app/data_providers/snowplow/subject.py

mmiermans:
@jsrawan-mobo Those references to user.user_id are unrelated to storing user preferences, and don't need to be changed/removed.

jsrawan-mobo (approved):

jsrawan-mobo:
> Thanks for searching through the repo, though! I always try to do the same with removal PRs.

I did see that the 'user_id' field wasn't removed so I suppose its still used for analytics.

Just to clarify - if you delete the 'old' feature group in Athena, then how does 'user_id' get populated in the RequestUser?

mmiermans:
> Just to clarify - if you delete the 'old' feature group in Athena, then how does 'user_id' get populated in the RequestUser?

@jsrawan-mobo For logged-in users, we receive `user_id` as a header from Apollo. A list of all headers is at the bottom of the [Pocket Graph readme page in Apollo Studio](https://studio.apollographql.com/graph/pocket-client-api/variant/current/home).

PR #1061, opened 2023-08-09
chore: remove unused Firefox recommendation code

mmiermans:
# Goal
Remove unused code related to our attempt in 2022 to migrate NewTab recommendations, because this could cause confusion later.

## Search for external dependencies
- There are no references to the removed slate ids outside of this repo and Pocket/firefox-newtab-proxy.
- [Pocket/firefox-newtab-proxy](https://github.com/Pocket/firefox-newtab-proxy) is archived and not deployed in production.
- There are no other references to the firefox-new-tab-engagement Feature Groups.

## QA
Before merging this PR search the [RecommendationAPI CloudWatch logs](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fecs$252FRecommendationAPI-Prod$252Fapp20201216004528169500000001) for the to-be-removed slate ids:
- [x] `f99178fb-6bd0-4fa1-8109-cda181b697f6` is not present in logs
- [x] `f4c3076c-2184-44ac-9272-0e7387f88fa6` is not present in logs
- [x] `d9b55d46-f909-43a2-876c-30784e90093a` is not present in logs
- [x] `79655eb2-47a1-4a26-9235-29e4768ff0a1` is not present in logs

## Deployment
- [ ] Delete firefox-new-tab-engagement Feature Groups

## Reference

Tickets:
* https://mozilla-hub.atlassian.net/browse/MC-27

PR #1062, opened 2023-08-11
chore: [MC-28] log slate_id and slate_lineup_id

mmiermans:
# Goal
In https://github.com/Pocket/recommendation-api/pull/1061, @jeshuaborges asked to be completely confident there are no GraphQL queries to `getSlate` for any `slate_id` values that will be removed in that PR.

## Deployment
After merging this PR, let it run for a few days, then search the [RecommendationAPI CloudWatch logs](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fecs$252FRecommendationAPI-Prod$252Fapp20201216004528169500000001) for the to-be-removed slate ids:

- [ ] `f99178fb-6bd0-4fa1-8109-cda181b697f6` is not present in logs
- [ ] `f4c3076c-2184-44ac-9272-0e7387f88fa6` is not present in logs
- [ ] `d9b55d46-f909-43a2-876c-30784e90093a` is not present in logs
- [ ] `79655eb2-47a1-4a26-9235-29e4768ff0a1` is not present in logs

## References
https://mozilla-hub.atlassian.net/browse/MC-28

## Implementation Decisions
After looking into [passing GraphQL variables](https://www.apollographql.com/docs/apollo-server/api/plugin/usage-reporting/#valid-sendvariablevalues-object-signatures) to Apollo, I decided that it would be easier to just add some logging statements.

By default Apollo doesn't report variables because of security risk. I'm not sure if we use variables for sensitive data or not. We could enable it selectively in [our Apollo Gateway](https://github.com/Pocket/client-api/blob/main/src/server/server.ts#L35-L38). In the end that road didn't feel worth the effort to me, but I can pursue it further if anyone feels that it's worth it in the long run.


rolf-moz:
I reviewed the usage logs in apollo studio and GetSlateLineup is called about once per second, so it shouldn't add too much to the logs. 
Do you want to add a Todo to remove this log message later do you think it can be left around?

rolf-moz (approved):
(comment left earlier)

mmiermans:
>  Do you want to add a Todo to remove this log message later do you think it can be left around?

No, I think this is useful logging to keep around for the future.

PR #1064, opened 2023-08-16
feat(mozsocialSlate): [MC-33] add demo endpoint for mozsocial slate

Herraj:
# Goal
Add a test/demo endpoint for any future MozSocial client to consume and request new tab recommendations data.

## Todos
- [ ] Update query name and other parameters after getting more clarity

Tickets:
* https://mozilla-hub.atlassian.net/browse/MC-33

## Implementation Decisions
Since we are just setting up a demo endpoint which returns the same query results as the new tab slate query, we've copied the existing implementation.

### DIFF
--- a/app/graphql/graphql_router.py
+++ b/app/graphql/graphql_router.py
@@ -37,6 +37,12 @@ class Query:
                     'The locale argument determines the UI and recommendation content language.'
     )
 
+    moz_social_slate: CorpusSlate = strawberry.field(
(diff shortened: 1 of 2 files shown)


mmiermans (approved):
We could not remember if there is a convention to mark GraphQL queries as 'in development', so we've added a string that should hopefully make clear that this endpoint is not meant to be used in production until we've had more discussion.

PR #1066, opened 2023-10-11
fix(Telemetry): Updated opentelemetry packages

mmiermans:
# Goal
Fix [dependabot/77](https://github.com/Pocket/recommendation-api/security/dependabot/77).

# How to test

## Test setup
- [x] `git push -f origin update-telemetry:dev`

## Test steps
- [x] Connect to the Pocket-Dev VPN
- [x] Make 50 requests for recommendations (sample rate is 10%)
```
for i in `seq 1 50`; do curl --location 'https://recommendation-api.getpocket.dev' --header 'Content-Type: application/json' --data '{
    "query": "query NewTabSlate($locale: String!, $region: String, $count: Int) { newTabSlate(locale: $locale, region: $region) { id recommendations(count: $count) { id tileId corpusItem { id } } }}",
    "variables": {"locale": "fr-FR", "region": "FR", "count": 30},
    "operationName": "NewTabSlate"
}'; done
```
- [x] Searching for [RecommendationAPI in X-Ray](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#xray:traces/query?~(query~(expression~'service*28id*28name*3a*20*22RecommendationAPI*22*20*2c*20type*3a*20*22AWS*3a*3aECS*3a*3aContainer*22*20*29*29)~context~())) should result in about 5 traces
- [x] Clinking into one of the traces show RecommendationAPI segments:

image

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -47,8 +47,8 @@ opentelemetry-api = "*"
 opentelemetry-sdk = "*"
 opentelemetry-sdk-extension-aws = "*"
 opentelemetry-exporter-otlp-proto-grpc = "*"
-opentelemetry-instrumentation-aiohttp-client = "==0.37b0"
-opentelemetry-instrumentation-fastapi = "==0.37b0"
+opentelemetry-instrumentation-aiohttp-client = "==0.41b0"
+opentelemetry-instrumentation-fastapi = "==0.41b0"
 
 [requires]
 python_version = "3.8"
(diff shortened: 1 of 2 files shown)


rolf-moz (approved):
Looks good. Any reason why we do exact matches on version, since the lock file fixes the version?

mmiermans:
> Looks good. Any reason why we do exact matches on version, since the lock file fixes the version?

@rolf-moz The reasoning was that this package is in beta, and is not expected to maintain backward compatibility. Because this is a telemetry package, breaking changes won't affect functionality, so I could go either way on this.

PR #1078, opened 2024-03-29
MC-767: Remove Hybrid CF experiment from Pocket Home

mmiermans:
# Goal
Remove the Hybrid Collaborative Filtering experiment from Pocket Home.

The Metaflow job powering this experiment was not functioning, and has been removed on March 1st: https://github.com/Pocket/dl-metaflow-jobs/pull/293

## QA

This change has been pushed to dev for QA.

1. Enable the Pocket-Dev VPN
2. Send the following request:
```shell
curl --location 'https://recommendation-api.getpocket.dev' \
--header 'userId: 342' \
--header 'encodedId: 234' \
--header 'consumerKey: 123-456-678' \
--header 'apiId: 123' \
--header 'applicationName: Postman' \
--header 'Jwt: 1234567679065906570965' \
--header 'Content-Type: application/json' \
--data '{
    "query": "query PostmanHome { homeSlateLineup(locale: \"en-US\") { slates { headline recommendations(count: 3) { id corpusItem { id } } } }}",
    "variables": null,
    "operationName": "PostmanHome"
}'
```

Expectations:
- [x] Response is successful and contains recommendations

## Reference

Tickets:
* https://mozilla-hub.atlassian.net/browse/MC-767

PR #1079, opened 2024-04-04
feat(v3-home): adding in a v3 home

bassrock:
# Goal

Pocket would like to experiment with a slightly different home experience and provide it to logged out users.

This does the following:
* Creates a feature flag for v3 home
* Enables home to return defaults instead of errors if there is no User passed through, enabling logged out.
* If the flag is enabled
    * Removes thompson sampling
    *  returns slates in a different order
    * removes topic slates
    * adds a pocket worthy slate
    * removes collections slate

I was able to test this on dev with the following query:
```graphql
query Home($locale: String!) {
  homeSlateLineup(locale: $locale) {
    id
    slates {
      headline
      id
      recommendations {
        reason {
          type
          name
        }
        id
      }
    }
  }
}
```

## Todos
- [x] Deploy https://github.com/Pocket/data-flows/pull/274 // can happen in a follow up deploy, the code will use collections slate for now.


app/data_providers/dispatch.py
```diff
@@ -171,39 +175,51 @@ async def get_slate_lineup(
...
             self.user_impression_cap_provider.get(user),
             self._get_preferred_topics(user))
 
         slates = []
-        if preferred_topics:
-            slates += [self.for_you_slate_provider.get_slate(
-                preferred_topics=preferred_topics,
-                user_impression_capped_list=user_impression_capped_list
-            )]
+        assignment = await self.unleash_provider.get_assignment(POCKET_HOME_V3_FEATURE_FLAG, user=user)
```
mmiermans, line 202: If you put this in the above `gather` call it will happen asynchronously with fetching the topics and impression caps. For example: https://github.com/Pocket/recommendation-api/pull/1078/files#diff-acaa726ac78ef290df20b378dad8f436fdb4f3d21b675a0ba561f76a03c0ed3cL198

app/data_providers/slate_providers/recommended_reads_slate_provider.py
```diff
@@ -23,6 +24,10 @@ async def rank_corpus_items(
             *args,
             **kwargs,
     ) -> List[CorpusItemModel]:
+        assignment = await self.unleash_provider.get_assignment(POCKET_HOME_V3_FEATURE_FLAG)
+        if assignment is not None and assignment.assigned is True:
+            return items
```
mmiermans, line 29: Why do you check the assignment here? Isn't that already covered by the code in dispatch.py?

app/data_providers/slate_providers/topic_slate_provider.py
```diff
@@ -40,5 +41,10 @@ async def rank_corpus_items(self, items: List[CorpusItemModel], *args, **kwargs)
         :param items: Candidate corpus items
         :return: Randomizes items.
         """
+        assignment = await self.unleash_provider.get_assignment(POCKET_HOME_V3_FEATURE_FLAG)
+        if assignment is not None and assignment.assigned is True:
+            # do not sample if they are in the v3 flag
+            return items
```
mmiermans, line 47: Same question here.

mmiermans (commented):
:muscle: Strong come-back into this repo!

app/data_providers/slate_providers/recommended_reads_slate_provider.py
bassrock: We check it here so that we can choose to not thompson sample within the slate

bassrock: Updated logic based on slack convo

mmiermans (approved):
🏆

PR #1083, opened 2024-04-11
feat(capping): adding impression capping for each slate that supports it

bassrock:
# Goal

I noticed home was not showing completely new content. Digging in I realized only the for you slate actually had user impression capping on.

## Todos
- [x] Add impression capping if we have a user object


mmiermans (approved):
We didn't enable this for other slates because we thought it wouldn't have much of an impact. The main reason why we enabled impression capping is because personalization has a tendency to repeatedly show the same content for some people.

Can't hurt to enable it across the board though. :+1:

PR #1089, opened 2024-05-23
fix: [MC-1139] flaky Thompson sampling test

mmiermans:
# Goal
[MC-1139](https://mozilla-hub.atlassian.net/browse/MC-1139) Fix a flaky test. `test_rank_by_ctr_over_n_trials` fails about 1% of the time:

> FAILED tests/unit/rankers test_algorithms.py::TestAlgorithmsThompsonSampling::test_rank_by_ctr_over_n_trials[thompson_sampling-metrics0] - assert 2 != 2.0

Looking at the code, at first sight it seems very improbable that the average rank is an integer. My understanding is that asserting the rank is not an integer intends to assert that it is not always ranked the same. However, these are different, because the average of [1, 2, 2, 2, 3] is exactly 2.  It’s hard to estimate the probability of this happening, but when I ran the test 1000 times it failed about 10 times.

[MC-1139]: https://mozilla-hub.atlassian.net/browse/MC-1139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit/rankers/test_algorithms.py
```diff
@@ -295,10 +301,10 @@ def test_rank_by_ctr_over_n_trials(self, thompson_sampling_function, metrics, nt
...
 
         # ranks are not deterministic
-        assert int(ranks['999999']) != ranks['999999']
-        assert int(ranks['666666']) != ranks['666666']
-        assert int(ranks['333333']) != ranks['333333']
-        assert int(ranks['222222']) != ranks['222222']
+        assert min_ranks['999999'] != max_ranks['999999']
+        assert min_ranks['666666'] != max_ranks['666666']
+        assert min_ranks['333333'] != max_ranks['333333']
+        assert min_ranks['222222'] != max_ranks['222222']
```
mmiermans, line 307: Asserting that the minimum rank is different from the maximum I think is closer to the intention. Even this failed occasionally, but by increasing ntrails the test became stable.

tests/unit/rankers/test_algorithms.py
```diff
@@ -254,13 +254,14 @@ def test_invalid_trailing_period(self):
                 trailing_period=123  # Model does not have 123 day trailing metrics
             )
 
+    @pytest.mark.parametrize("repeat", range(10))  # Increase 100x when changing thompson_sampling to avoid a flaky test
```
mmiermans, line 257: Repeating this test 10 times might help to distinguish a flaky test from an actual bug, if it fails 1/10 times. We also have an `ntrials` parameter. I think they serve different purposes?

lfishhead (approved):

PR #1096, opened 2024-08-21
feat(syndication): removing syndication as a test

bassrock:
# Goal

Testing with an a/b test if/and how removing the syndication slate affects engagement

Flag: https://featureflags.readitlater.com/projects/default/features/temp.recommendation-api.pocket-home-no-syndication

### DIFF
--- a/app/data_providers/dispatch.py
+++ b/app/data_providers/dispatch.py
@@ -6,7 +6,8 @@ from asyncio import gather
 from typing import List, Coroutine, Any, Tuple, Optional
 from datetime import datetime
 
-from app.config import DEFAULT_TOPICS, GERMAN_HOME_TOPICS, POCKET_HOME_PRIDE_FEATURE_FLAG, POCKET_HOME_V4_FEATURE_FLAG
+from app.config import DEFAULT_TOPICS, GERMAN_HOME_TOPICS, \
+    POCKET_HOME_NO_SYNDICATION_FEATURE_FLAG
 from app.data_providers.corpus.corpus_feature_group_client import CorpusFeatureGroupClient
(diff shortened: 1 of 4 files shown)


app/data_providers/dispatch.py
```diff
@@ -214,6 +217,11 @@ async def get_en_us_slate_lineup(
             self.life_hacks_slate_provider.get_slate(enable_thompson_sampling_with_seed=seed_id,
                                                      user_impression_capped_list=user_impression_capped_list),
         ]
+        if no_syndication_assignment is not None:
+            if no_syndication_assignment.assigned is True:
+                del slates[0]
+            asyncio.create_task(
+                self.snowplow.track_variant_enroll(assignment=no_syndication_assignment, user=user, api_client=api_client))
```
mmiermans, line 224: This event is [fired off automatically](https://github.com/Pocket/recommendation-api/blob/3e8aee030c4812d1d09ef97999034580b8c4f774/app/data_providers/snowplow/snowplow_corpus_recommendations_tracker.py#L39) if you return the assignment on line 231. Here's [an example of how we did this before](https://github.com/Pocket/recommendation-api/pull/1078/files#diff-acaa726ac78ef290df20b378dad8f436fdb4f3d21b675a0ba561f76a03c0ed3cL232). This will ensure we know the assignment for each individual click and impression event, which I believe the [Home Experiment Report](https://app.mode.com/getpocket/reports/955892efcbf4) relies on.

bassrock: updated!

mmiermans (approved):
Good except for the indentation error.

PR #1098, opened 2024-08-22
fix(locales): updating the config to use the english ids for mapping

bassrock:
# Goal

It turns out when passing topic ids to the topic slate provider, you should be using the english ids, which then get merged with metadata from dynamodb, and our translations and then re-mapped into their real candidate set ids from the _TOPIC_CANDIDATE_SETS variables

I tested this locally against Prod.

This also puts all the extra locales behind an experiment feature flag.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -124,3 +124,4 @@ ES_ES_HOME_TOPICS = [
 
 # Changes the slate lineups for home and removes syndication as a a/b test
 POCKET_HOME_NO_SYNDICATION_FEATURE_FLAG = 'temp.recommendation-api.pocket-home-no-syndication'
+POCKET_HOME_MORE_LOCALES = 'temp.recommendation-api.home.more-locales'
--- a/app/config.py
+++ b/app/config.py
@@ -75,6 +75,10 @@ qa_slate_map = {
 # For running Open Telemetry locally, set `OTEL_DAEMON_ADDRESS`, otherwise defaults to production value.
 otel_daemon_address = os.getenv('OTEL_DAEMON_ADDRESS', 'http://127.0.0.1:4317')
 
+# Note: All the topics here use the English version of the Candidate Set Ids.
+# They then get merged with the metadata from DynamoDB, TopicMetadata table and our translation files in this repository.
+# From there it is then mapped to the real candidate set id for that language from the _TOPIC_CANDIDATE_SETS variable based on the locale model
+
 DEFAULT_TOPICS = [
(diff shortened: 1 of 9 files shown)


app/config.py
```diff
@@ -75,6 +75,10 @@
 # For running Open Telemetry locally, set `OTEL_DAEMON_ADDRESS`, otherwise defaults to production value.
 otel_daemon_address = os.getenv('OTEL_DAEMON_ADDRESS', 'http://127.0.0.1:4317')
 
+# Note: All the topics here use the English version of the Candidate Set Ids.
+# They then get merged with the metadata from DynamoDB, TopicMetadata table and our translation files in this repository.
+# From there it is then mapped to the real candidate set id for that language from the _TOPIC_CANDIDATE_SETS variable based on the locale model
```
mmiermans, line 80: It's been a while, so I had to dive in here again. These are topic ids, not candidate set ids.

I guess at some point we started transitioning from a UUID to a more readable "corpus_topic_id". We should use the latter for readability, and refactor [topic_provider](https://github.com/Pocket/recommendation-api/blob/7e12258c8203214b68ca031fc925914ed8d75636/app/data_providers/topic_provider.py#L45-L52) to fetch a topic using for example "BUSINESS".

app/config.py
```diff
@@ -91,32 +95,33 @@
...
 
 ES_ES_HOME_TOPICS = [
-    '177d3d8d-7733-402d-bf76-39eb9c2e24cd',  # Technology
-    'e6e1696e-571a-49a6-ac27-1de508fc1f31',  # Business
-    'bd983eba-0b4a-4916-82dd-d3a212ce934b',  # Science
-    'b3c241b0-3651-4460-977c-568cbfbef2c3',  # Self-improvement
+    '25c716f1-e1b2-43db-bf52-1a5553d9fb74',  # Technology
+    '1bf756c0-632f-49e8-9cce-324f38f4cc71',  # Business
+    '058011b8-c70d-4a25-92e5-478e3ff0f0e6',  # Science
+    '45f8e740-42e0-4f54-8363-21310a084f1f',  # Self-improvement
```
mmiermans, line 122: Given that these are all the same values, a lot of duplicate code could be removed:
- Create `NON_EN_US_HOME_TOPICS`
- Collapse `get_xx_yy_slate_lineup`. I think these are all just duplicates? I should have suggested that before.

mmiermans (approved):
👍 Glad we got to tackle some tech debt!

PR #1110, opened 2024-10-28
(chore): MC-511: upgrade lambda python

katerinachinnappan:
# Goal

Python 3.8 has been deprecated starting October 14.Upgrading the Python version to Python3.9 for the following lambda functions:

* `RecommendationAPI-Prod-SQS-Translation`
* `cwsyn-recapi`

Decided not to upgrade to the most recent pocket tf modules (v5) as this introduced a lot of breaking changes, so referencing `python3.9` lambda runtime directly. According to [AWS](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html), there is no scheduled date to deprecate this version (as of now). 

`python3.9` seems to be the best version to use for now. Tried upgrading first to `Python3.12`, but there [has been an issue during build](https://stackoverflow.com/questions/77274572/multiqc-modulenotfounderror-no-module-named-imp). Upgrading to `python3.10` & `python3.11` also was problematic due to `cdktf` [dependency issues ](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/410318598490/projects/RecommendationAPI-Dev/build/RecommendationAPI-Dev%3Adbed8f2f-3766-45dc-920b-378115add03b/?region=us-east-1). 

## Todos
- [x] deployed to dev

## Reference

Tickets:
* https://mozilla-hub.atlassian.net/browse/MC-1511

### DIFF
--- a/.aws/src/monitoring.ts
+++ b/.aws/src/monitoring.ts
@@ -167,7 +167,7 @@ export class RecommendationApiSynthetics extends Construct {
               'RECOMMENDATION_API_DOMAIN': config.domain,
           }
         },
-        runtimeVersion: 'syn-python-selenium-1.3',
+        runtimeVersion: 'syn-python-selenium-4.1',
         schedule: {
           expression: 'rate(5 minutes)', // run every 5 minutes
         },
--- a/aws_lambda/tests/test_sqs_lambda.py
+++ b/aws_lambda/tests/test_sqs_lambda.py
@@ -7,7 +7,7 @@ import time
 import pytest
 from mypy_boto3_dynamodb.service_resource import DynamoDBServiceResource
 import boto3
-from moto import mock_dynamodb2
+from moto import mock_aws
 
 from aws_lambda.tests.fixtures.lambda_sqs_event import event, body, body2
 from aws_lambda.tests.fixtures.lambda_sqs_event_without_id import event_without_id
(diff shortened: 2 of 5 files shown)


.aws/src/monitoring.ts
```diff
@@ -167,7 +167,7 @@ export class RecommendationApiSynthetics extends Construct {
               'RECOMMENDATION_API_DOMAIN': config.domain,
           }
         },
-        runtimeVersion: 'syn-python-selenium-1.3',
+        runtimeVersion: 'syn-python-selenium-4.1',
```
katerinachinnappan, line 170: 1.3 has been deprecated back in March. https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Library_python_selenium.html

aws_lambda/tests/test_sqs_lambda.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 from mypy_boto3_dynamodb.service_resource import DynamoDBServiceResource
 import boto3
-from moto import mock_dynamodb2
+from moto import mock_aws
```
katerinachinnappan, line 10: latest release of `moto` removes all decorators with `mock_aws`

mmiermans (approved):
Good choice to move to Python 3.9, instead of investing additional effort in code that may very well not be around in a year! 👍

